PR #121, opened 2014-03-04
Tab complete file paths containing spaces

### LINKED ISSUE #83: Tab-completing filenames with spaces
Filenames with spaces don't appear to be handled correctly for tab completion.

They are completed, but the space is not escaped and the file path is not in quotes.

georgebrock:
- Adds support for tab completing file names with spaces.
- Handles quoted arguments correctly (i.e. spaces are not escaped within quotes)

Fixes #83

### DIFF
--- a/lib/gitsh/completer.rb
+++ b/lib/gitsh/completer.rb
@@ -36,14 +36,6 @@ module Gitsh
         end
       end
 
-      def full_input
-        readline.line_buffer
-      end
-
-      def input_before_current_argument
-        full_input[0...-input.length]
-      end
-
       def escape(arg)
         if completing_quoted_argument?
           arg.strip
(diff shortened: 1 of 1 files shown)


lib/gitsh/completer.rb
```diff
@@ -37,11 +37,30 @@ def available_completers
       end
 
       def completing_arguments?
-        full_input = readline.line_buffer
         tokens = full_input.split
         tokens.any? && full_input.end_with?(' ') || tokens.size > 1
       end
 
+      def completing_quoted_argument?
```
georgebrock: This could probably be more accurately implemented using `rl_completion_found_quote`, `rl_completion_quote_character`, or some combination of the two.

Issue #198, opened 2014-11-10
Support nested subshells


georgebrock:
The current parsing rules for subshells are a little naive: the subshell command will be everything from the `$(` to the first `)`, regardless of any other opening parentheses in between.

This prevent nested subshells, and also subshells containing parentheses for any other reason.

georgebrock:
Subshells are now merged, so we should get this done before the next release.

georgebrock:
Closed in favour of #218

PR #199, opened 2014-11-10
Subshells

georgebrock:
Support for subshells, e.g. `:echo "In $(!pwd)"`

The current parser rules don't support nesting, which might be a blocker (see #198), but I wanted to open a PR with this before it got too huge.

(Specs pass locally, Travis is suffering from the usual Readline segfault shenanigans)


georgebrock:
Thanks for the review, Joe. Finally merging this now that 0.9 is out; I'll prioritise fixing nested subshells for the next release.

PR #208, opened 2014-12-26
Fix install download URL

lf-:


georgebrock:
@lf- Thanks for the fix. The build failure is an unrelated Ruby/readline problem which I've not managed to track down yet.

I've merged this patch as bbb53ad031c95d78a364c090fc8620c645aeac62 (the commit ID changed because I re-wrote the commit message to describe the fix)

Issue #197, opened 2014-11-10
Variables in arguments should be lazily evaluated


georgebrock:
Variable values are substituted into commands by the `Gitsh::Transformer` as part of the parsing process. The result is that a `Gitsh::Command::Tree::*`—which composes two or more commands—doesn't work as expected when the LHS modified a variable that is used on the RHS.

Lazily evaluating argument values as part of the `Gitsh::Command::*#execute` method should solve this problem.

The expected behaviour is:

```
% :set x "1"
% :set x "2"; :echo $x
2
% :echo $x
2
```

The actual behaviour is:

```
% :set x "1"
% :set x "2"; :echo $x
1
% :echo $x
2
```

georgebrock:
Now that unset variables are going to be an error, when they're lazily evaluated it should also be possible to handle this kind of situation:

```
% :set b "This is B"
% :echo $a || :echo $b
This is B
```

georgebrock:
This is more-or-less done on the `gb-lazy-variables-original` branch. Waiting on #210 to open a PR.

Issue #136, opened 2014-03-12
Unset variables should error


mike-burns:
`commit $foo`, when `$foo` is unset, should print an error message and not continue.

This is in `transformer.rb`, in the `:var` rule.

Currently it passes an empty string to `#spawn` as the last argument:

    ["/usr/bin/env", "git", "commit"@0, ""]

Ideally this should do something more predictable: either error, or pass the empty string (`"\"\""`). Calle and I vote: error.

georgebrock:
I was wrong. Unset variables should be errors.

PR #209, opened 2015-01-09
Eliminate nils

georgebrock:
This PR paves the way for making unset variables raise an error (#136) which in turn will pave the way for lazily evaluating variables in command arguments (#197).

After attempting to implement lazy variable evaluation directly it became apparent that staying consistent with the way general purpose Unix shells handle unset variables is going to introduce a lot of complexity, far more than keeping that behaviour would justify (@mike-burns never liked it anyway—see discussion on #136—and I probably should have listened).

* Modify `Gitsh::Environment#fetch` and `Gitsh::GitRepository#config` to raise if no default value is given.
* Move default values from arguments to blocks to easily distinguish between “there is no default” and “the default is set to `nil`”.
* Remove `Gitsh::Environment#[]` and `Gitsh::MagicVariables#[]` entirely, in favour of `#fetch` methods.
* Update `Gitsh::Environment#fetch` to read from magic variables, which were previously only accessible via `#[]`.

### DIFF
--- a/lib/gitsh/environment.rb
+++ b/lib/gitsh/environment.rb
@@ -20,7 +20,7 @@ module Gitsh
       if force_default
         DEFAULT_GIT_COMMAND
       else
-        fetch('gitsh.gitCommand', DEFAULT_GIT_COMMAND, true)
+        fetch('gitsh.gitCommand', true) { DEFAULT_GIT_COMMAND }
       end
     end
 
@@ -28,19 +28,18 @@ module Gitsh
       self['gitsh.gitCommand'] = git_command
     end
 
-    def [](key)
-      magic_variables[key.to_sym] || variables[key.to_sym] || repo.config(key.to_s)
-    end
-
     def []=(key, value)
       variables[key.to_sym] = value
     end
 
-    def fetch(key, default, force_default_git_command = false)
-      variables.fetch(
-        key.to_sym,
-        repo.config(key.to_s, default, force_default_git_command)
-      )
+    def fetch(key, force_default_git_command = false, &block)
+      magic_variables.fetch(key.to_sym) do
+        variables.fetch(key.to_sym) do
+          repo.config(key.to_s, force_default_git_command, &block)
+        end
+      end
+    rescue KeyError
+      raise KeyError, "Variable '#{key}' is not set"
     end
 
     def config_variables
@@ -88,7 +87,7 @@ module Gitsh
     end
 
     def repo_config_color(name, default)
-      if color_override = self[name]
+      if color_override = fetch(name) { false }
         repo.color(color_override)
       else
         repo.config_color(name, default)
--- a/lib/gitsh/git_repository.rb
+++ b/lib/gitsh/git_repository.rb
@@ -54,7 +54,7 @@ module Gitsh
       git_output('remote').lines
     end
 
-    def config(name, default = nil, force_default_git_command = false)
+    def config(name, force_default_git_command = false)
       command = git_command(
         "config --get #{Shellwords.escape(name)}",
         force_default_git_command
@@ -62,8 +62,10 @@ module Gitsh
       out, err, status = Open3.capture3(command)
       if status.success?
         out.chomp
+      elsif block_given?
+        yield
       else
-        default
+        raise KeyError, "Git configuration variable #{name} is not set"
       end
     end
 
--- a/lib/gitsh/magic_variables.rb
+++ b/lib/gitsh/magic_variables.rb
@@ -4,9 +4,11 @@ module Gitsh
       @repo = repo
     end
 
-    def [](key)
+    def fetch(key)
       if available_variables.include?(key)
         send(key)
+      else
+        yield
       end
     end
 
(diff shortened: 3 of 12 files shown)


lib/gitsh/magic_variables.rb
```diff
@@ -4,9 +4,13 @@ def initialize(repo)
       @repo = repo
     end
 
-    def [](key)
+    def fetch(key)
       if available_variables.include?(key)
         send(key)
+      elsif block_given?
```
jferris: It looks like this is only used from `Environment`, and a block is always provided.

georgebrock: Good point, I'll remove the extra branch from the conditional.

lib/gitsh/environment.rb
```diff
@@ -88,7 +87,7 @@ def repo_has_untracked_files?
     end
 
     def repo_config_color(name, default)
-      if color_override = self[name]
+      if color_override = fetch(name) { nil }
```
jessieay: I haven't seen this syntax with `fetch` before - what is the purpose of passing `{ nil }` here?

georgebrock: If the key doesn't exist, `#fetch` will yield to the block for the default value. In this case, the default value is `nil`.

lib/gitsh/git_repository.rb
```diff
@@ -62,8 +62,10 @@ def config(name, default = nil, force_default_git_command = false)
       out, err, status = Open3.capture3(command)
       if status.success?
         out.chomp
+      elsif block_given?
+        yield
       else
-        default
+        raise KeyError, "Git configuration variable #{name} is not set"
```
jessieay: I am wondering if these changes make sense in this case - if we are going to raise an error when the config var isn't set, why not just call it normally rather than `fetch`? Is it just so the error message is more readable?

lib/gitsh/environment.rb
jessieay: Interesting - so how is this better than what we had before? (I previously understood the benefit of `fetch` to be that it would not return `nil`)

jessieay: oh wait, I think I may have misunderstood. This is if the KEY doesn't exist, not if it doesn't have a value. Nevermind then! :)

spec/units/git_repository_spec.rb
```diff
@@ -171,32 +171,40 @@
...
-          run 'git config --local alias.zecho "!echo zzz"'
-          expect(repo.config('alias.zecho')).to eq '!echo zzz'
+    context 'for a variable that is set' do
+      it 'returns a git configuration value' do
+        with_a_temporary_home_directory do
+          in_a_temporary_directory do
+            repo = Gitsh::GitRepository.new(env)
+            run 'git init'
+            run 'git config --local alias.zecho "!echo zzz"'
+            expect(repo.config('alias.zecho')).to eq '!echo zzz'
```
jessieay: newlines to separate spec phases?

lib/gitsh/environment.rb
georgebrock: I removed the `#[]` method, to avoid having a method that might return `nil`: This is one of the very few cases where a `nil` would be useful, and I thought it was better to be explicit and use the slightly odd looking `{ nil }` here, rather than keep the `#[]` method that might let `nil` values accidentally leak into places where they're not wanted.

spec/units/git_repository_spec.rb
georgebrock: Good call. Done.

spec/units/git_repository_spec.rb
```diff
@@ -171,32 +171,43 @@
...
-          run 'git init'
-          run 'git config --local alias.zecho "!echo zzz"'
-          expect(repo.config('alias.zecho')).to eq '!echo zzz'
+    context 'for a variable that is set' do
+      it 'returns a git configuration value' do
+        with_a_temporary_home_directory do
+          in_a_temporary_directory do
+            repo = Gitsh::GitRepository.new(env)
+            run 'git init'
+            run 'git config --local alias.zecho "!echo zzz"'
```
jessieay: would be helpful to make `"!echo zzz"` a variable here - would improve readability of test, in my opinion!

lib/gitsh/git_repository.rb
georgebrock: Yes, it's mostly to make the message clear: this is where the exception comes from originally with a message indicating that the variable we're looking for didn't match any Git configuration variables, but it's re-raised in `Environment#fetch` with a different message to make it clear that the variable didn't exist anywhere we looked for it.

spec/units/git_repository_spec.rb
georgebrock: Good call. Extracted.

georgebrock:
Spec failure is unrelated (it's a Ruby/readline/threading bug)

lib/gitsh/environment.rb
mike-burns: You want `false`, not `nil`, because it's a Boolean test (`if`).

georgebrock: Oh, yes, good call.

georgebrock:
Thanks for the reviews, all.

PR #210, opened 2015-01-09
Consider unset variables to be an error.

### LINKED ISSUE #136: Unset variables should error

georgebrock:
If a command uses a variable that has not previously been set, the command will not be executed, and an error message will be printed identifying the unset variable.

Fixes #136

### DIFF
--- a/lib/gitsh/environment.rb
+++ b/lib/gitsh/environment.rb
@@ -1,3 +1,4 @@
+require 'gitsh/error'
 require 'gitsh/git_repository'
 require 'gitsh/magic_variables'
 
@@ -39,7 +40,7 @@ module Gitsh
         end
       end
     rescue KeyError
-      raise KeyError, "Variable '#{key}' is not set"
+      raise Gitsh::UnsetVariableError, "Variable '#{key}' is not set"
     end
 
     def config_variables
--- a/man/man1/gitsh.1
+++ b/man/man1/gitsh.1
@@ -239,13 +239,10 @@ branch.
 .El
 .El
 .Pp
-Unset variables will be ignored, as then are in
-.Xr sh 1 .
-For example, the commit message produced by the following command will be
-.Ic "a commit" :
-.Bd -literal -offset indent
-@ commit -m $unset "a commit"
-.Ed
+Attempting to use an unset variable will cause a command to fail. This is
+different to
+.Xr sh 1 ,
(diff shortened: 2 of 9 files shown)


man/man1/gitsh.1
```diff
@@ -239,13 +239,10 @@ branch.
...
 .Pp
-Unset variables will be ignored, as then are in
-.Xr sh 1 .
-For example, the commit message produced by the following command will be
-.Ic "a commit" :
-.Bd -literal -offset indent
-@ commit -m $unset "a commit"
-.Ed
+Attempting to use an unset variable will cause a command to fail. This is
+different to
```
mike-burns: I had to look this up: `different to` is en-GB and `different than` is en-US. Now we know.

lib/gitsh/environment.rb
```diff
@@ -39,7 +40,7 @@ def fetch(key, force_default_git_command = false, &block)
         end
       end
     rescue KeyError
-      raise KeyError, "Variable '#{key}' is not set"
+      raise Gitsh::Error, "Variable '#{key}' is not set"
```
mike-burns: It often proves useful to know what kind of exception happened without string parsing. This is the first exception class, so it doesn't matter yet, but how about starting off with a `Gitsh::KeyError` subclass here?

man/man1/gitsh.1
georgebrock: Huh, I didn't know that. I'm learning a lot of en-US this week: I didn't know “donut” was the en-US spelling of “doughnut” until yesterday.

I'm happy to leave this as is, I think the phrase is similar enough to be comprehensible.

lib/gitsh/environment.rb
georgebrock: Good call. I've called it `Gitsh::UnsetVariableError` instead, because I think that better captures the purpose.

georgebrock:
Thanks for the review, @mike-burns.

PR #211, opened 2015-01-11
Lazily evaluate variables

### LINKED ISSUE #197: Variables in arguments should be lazily evaluated

georgebrock:
Originally, gitsh would replace a variable with its value when transforming a command's AST into a command object. This made it impossible to write and then read a variable in the same command, e.g. with `:set x 1 && :echo $x`

Fixes #197

### DIFF
--- a/lib/gitsh/argument_list.rb
+++ b/lib/gitsh/argument_list.rb
@@ -1,0 +1,21 @@
+module Gitsh
+  class ArgumentList
+    def initialize(args)
+      @args = args
+    end
+
+    def length
+      args.length
+    end
+
+    def values(env)
+      @args.map do |arg|
+        arg.value(env)
+      end
+    end
+
+    private
+
+    attr_reader :args
+  end
+end
--- a/lib/gitsh/argument_builder.rb
+++ b/lib/gitsh/argument_builder.rb
@@ -1,0 +1,52 @@
+require 'gitsh/arguments/composite_argument'
+require 'gitsh/arguments/string_argument'
+require 'gitsh/arguments/variable_argument'
+
+module Gitsh
+  class ArgumentBuilder
+    def self.build
+      builder = new
+      yield builder
+      builder.argument
+    end
+
+    def initialize
+      @arguments = []
+      @literals = []
+    end
+
+    def add_literal(literal)
+      literals << literal
+    end
+
+    def add_variable(variable_name)
+      collect_literals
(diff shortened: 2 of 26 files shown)


mike-burns:
I like the elegance.

This is repeated, which is fine:

    args.map { |arg| arg.value(env) }

But it makes me wonder whether `env` should just know about `args`, or whether there should be an object that combines `env` and `args`.

georgebrock:
Thanks!

I was considering a `Gitsh::ArgumentList` to wrap up the conversion from arguments to values. I'll give that a go and see if it improves things.

georgebrock:
@mike-burns Paired with @sarony to add an `ArgumentList` class. Ready for another review.

lib/gitsh/argument_list.rb
```diff
@@ -0,0 +1,21 @@
+module Gitsh
+  class ArgumentList
+    def initialize(args)
+      @args = args
+    end
+
+    def length
+      args.length
+    end
```
mike-burns: Where is `#length` used? The only place I see it in this PR, aside from the unit tests, are in `ArgumentBuilder::combined_arguments` -- but that's for an Array created in the initializer, not for an ArgumentList.

mike-burns:
I like that new object. Just one question around `ArgumentList#length`; if I'm misreading something then it's good to merge.

lib/gitsh/argument_list.rb
georgebrock: It's used by the command classes to verify that their argument lists contain the correct number of arguments (e.g. [internal_command.rb:65](https://github.com/thoughtbot/gitsh/blob/5136aa62b1c2c6fd233bd97cac760b45de7326af/lib/gitsh/commands/internal_command.rb#L65)).

mike-burns:
Cool, looks good to merge.

georgebrock:
Thanks for the review.

Issue #114, opened 2014-02-26
Tests fail under GNU readline on Linux


mike-burns:
As mentioned in #109:

    % dpkg -s libreadline6-dev
    Package: libreadline6-dev
    Status: install ok installed
    Priority: optional
    Section: libdevel
    Installed-Size: 709
    Maintainer: Matthias Klose <doko@debian.org>
    Architecture: amd64
    Multi-Arch: same
    Source: readline6
    Version: 6.2+dfsg-0.1
    Depends: libreadline6 (= 6.2+dfsg-0.1), libtinfo-dev, dpkg (>= 1.15.4) | install-info
    Conflicts: libreadline-gplv2-dev, libreadline5-dev
    Description: GNU readline and history libraries, development files
     The GNU readline library aids in the consistency of user interface
     across discrete programs that need to provide a command line
     interface.
     .
     The GNU history library provides a consistent user interface for
     recalling lines of previously typed input.
     .
     This package contains development files and the readline documentation
     in info format.
    % ruby -v
    ruby 1.9.3p484 (2013-11-22 revision 43786) [x86_64-linux]
    % 
    % rspec
    ........................../home/mike/gitsh/spec/support/fake_readline.rb:16: [BUG] Segmentation fault
    ruby 2.0.0p353 (2013-11-22) [x86_64-linux-gnu]
    
    -- Control frame information -----------------------------------------------
    c:0009 p:---- s:0035 e:000034 CFUNC  :readline
    c:0008 p:0053 s:0030 e:000029 METHOD /home/mike/gitsh/spec/support/fake_readline.rb:16
    c:0007 p:0017 s:0025 e:000024 METHOD /home/mike/gitsh/lib/gitsh/readline_blank_filter.rb:5
    c:0006 p:0014 s:0020 e:000019 METHOD /home/mike/gitsh/lib/gitsh/cli.rb:64
    c:0005 p:0023 s:0016 e:000015 METHOD /home/mike/gitsh/lib/gitsh/cli.rb:54
    c:0004 p:0027 s:0012 e:000011 METHOD /home/mike/gitsh/lib/gitsh/cli.rb:43
    c:0003 p:0029 s:0009 e:000008 METHOD /home/mike/gitsh/lib/gitsh/cli.rb:31
    c:0002 p:0117 s:0006 e:000005 BLOCK  /home/mike/gitsh/spec/support/gitsh_runner.rb:46 [FINISH]
[...]

PR #212, opened 2015-01-12
Abbreviate branch name

### LINKED ISSUE #143: Truncate branch name in prompt
Some users have absurdly long branch names. We should elide those at, say, 15 chars.

georgebrock:
By default, only show the first 15 characters of the branch name in the prompt.

Fixes #143

### DIFF
--- a/lib/gitsh/prompter.rb
+++ b/lib/gitsh/prompter.rb
@@ -1,9 +1,12 @@
+# encoding: utf-8
+
 require 'gitsh/colors'
 require 'gitsh/prompt_color'
 
 module Gitsh
   class Prompter
-    DEFAULT_FORMAT = "%D %c%b%#%w".freeze
+    DEFAULT_FORMAT = "%D %c%B%#%w".freeze
+    BRANCH_CHAR_LIMIT = 15
 
     def initialize(options={})
       @env = options.fetch(:env)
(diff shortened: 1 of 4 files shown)


georgebrock:
(Spec failure is related to #114)

lib/gitsh/prompter.rb
```diff
@@ -35,6 +39,14 @@ def prompt_format
       env.fetch('gitsh.prompt') { DEFAULT_FORMAT }
     end
 
+    def shortened_branch_name
+      if branch_name.length > BRANCH_CHAR_LIMIT
+        branch_name[0...BRANCH_CHAR_LIMIT] + "…"
+      else
+        branch_name
+      end
```
mike-burns: Though this could be:

```ruby
def shortened_branch_name
  branch_name[0..BRANCH_CHAR_LIMIT] + ellipsis
end
def ellipsis
  if branch_name.length > BRANCH_CHAR_LIMIT
    "..."
  else
    ""
end
```

I think the above is more clear because the conditional has fewer things in it, but it might also be more ugly.

man/man1/gitsh.1
```diff
@@ -125,7 +126,8 @@ command. The following placeholders can be used in the custom prompt:
 .Bl -column ".Sy Placeholder" ".Sy Meaning" -offset indent
 .It Sy Placeholder  Ta Sy Meaning
 .It %#              Ta terminating sigil (see previous table)
-.It %b              Ta current HEAD (branch name, tag name, etc.)
+.It %b              Ta current HEAD (full branch name, tag name, etc.)
+.It \&%B            Ta current HEAD, limited to 15 characters
```
mike-burns: Nice. I like this solution.

lib/gitsh/prompter.rb
georgebrock: I like it. Done.

georgebrock:
Thanks for the review, merging.

PR #214, opened 2015-01-14
Command error handling

### LINKED ISSUE #213: Unset variable errors should be handled at the command level
Unset variable errors—and possibly other classes of error in future—are currently handled by the `Gitsh::Interpreter`, so that they kill the current command but not the whole gitsh session.

If this kind of error occurs on a line which contains multiple commands, the whole line will be aborted, but it should be possible to do things like this:

```
% :set b "This is B"
% :echo $a || :echo $b
This is B
```

Moving the error handling from the `Gitsh::Interpreter` to the `Gitsh::Commands::*` classes should help with this.

georgebrock:
This PR moves the error handling from the `Gitsh::Interpreter` to a `Gitsh::Commands::ErrorHandler` decorator, so that a line that contains multiple commands can continue, even if one of those commands fails.

It makes it possible to do things like this:

```
% :set b "This is B"
% :echo $a || :echo $b
This is B
```

Fixes #213

### DIFF
--- a/lib/gitsh/commands/error_handler.rb
+++ b/lib/gitsh/commands/error_handler.rb
@@ -1,0 +1,21 @@
+module Gitsh
+  module Commands
+    class ErrorHandler
+      def initialize(command, env)
+        @command = command
+        @env = env
+      end
+
+      def execute
+        command.execute
+      rescue Gitsh::Error => error
+        env.puts_error("gitsh: #{error.message}")
+        false
+      end
+
+      private
+
+      attr_reader :command, :env
+    end
+  end
+end
--- a/lib/gitsh/commands/factory.rb
+++ b/lib/gitsh/commands/factory.rb
@@ -1,0 +1,36 @@
+require 'gitsh/argument_list'
+
+module Gitsh
+  module Commands
+    class Factory
+      def initialize(command_class, context)
+        @command_class = command_class
+        @context = context
+      end 
+
+      def build
+        command_class.new(env, command, argument_list)
+      end
+
+      private
+
+      attr_reader :command_class, :context
+
+      def argument_list
+        ArgumentList.new(args)
+      end
+
+      def env
+        context[:env]
+      end
+
+      def command
+        context[:cmd]
+      end
+
+      def args
+        context.fetch(:args, []).compact
+      end
+    end
+  end
+end
--- a/lib/gitsh/commands/factory.rb
+++ b/lib/gitsh/commands/factory.rb
@@ -1,4 +1,5 @@
 require 'gitsh/argument_list'
+require 'gitsh/commands/error_handler'
 
 module Gitsh
   module Commands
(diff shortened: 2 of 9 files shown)


spec/units/commands/factory_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'spec_helper'
+require 'gitsh/commands/factory'
+
+describe Gitsh::Commands::Factory do
+  describe '#build' do
+    it 'returns an instance of the given command class' do
```
georgebrock: @hrs and I were both unhappy with this spec, but we weren't sure how to improve it.

lib/gitsh/commands/error_handler.rb
```diff
@@ -0,0 +1,21 @@
...
+  module Commands
+    class ErrorHandler
+      def initialize(command, env)
+        @command = command
+        @env = env
+      end
+
+      def execute
+        command.execute
+      rescue Gitsh::Error => e
```
gabebw: "error" instead of "e"?

lib/gitsh/commands/factory.rb
```diff
@@ -0,0 +1,41 @@
...
+      def argument_list
+        ArgumentList.new(args)
+      end
+
+      def env
+        context[:env]
+      end
+
+      def command
+        context[:cmd]
```
gabebw: Why not spell out "command"?

lib/gitsh/commands/error_handler.rb
georgebrock: Good call. Done.

spec/units/commands/error_handler_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+        expect(handler.execute).to be command_instance.execute
+      end
+    end
+
+    context 'the command raises an error' do
+      it 'prints the error and returns false' do
+        env = stub('env', puts_error: nil)
+        command_instance = stub('command_instance')
+        command_instance.stubs(:execute).raises(Gitsh::Error, 'Oh noes!')
```
gabebw: `allow(command_instance).to receive(:execute).and_raise(Gitsh::Error, 'Oh noes!')` is the RSpec 3.x way to do it, I believe

lib/gitsh/commands/factory.rb
georgebrock: Good call. Done.

spec/units/commands/error_handler_spec.rb
georgebrock: This project mocks with mocha for historical reasons (RSpec didn't have test spies when I started the project). You're the second or third person to be surprised by this, so I've opened an issue: #215

gabebw: Oh hey Mocha. `.stubs` vs `.stub`, always confusing me.

georgebrock:
@gabebw Thanks for the review, Gabe. Any other feedback, or are we good to merge?

gabebw:
Oops! :+1:

georgebrock:
Thanks, Gabe.

PR #216, opened 2015-01-15
Update INSTALL in the configure script.

georgebrock:
The way we were building the `INSTALL` instructions was causing `make distcheck` to fail: `make clean` wasn't cleaning up the `INSTALL` file, which was what we wanted for building in the source directory, but not when building from another directory.

Letting autoconf handle the `INSTALL` file fixes this problem, and simplifies the Makefile in the process.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,10 +76,8 @@ gitsh is packaged and installed using GNU autotools.
         ./autogen.sh
         ./configure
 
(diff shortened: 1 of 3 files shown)


georgebrock:
(Spec failure is related to #114)

mike-burns:
Good catch. OK by me.

georgebrock:
Thanks, Mike. Merging.

PR #217, opened 2015-01-16
:source command

georgebrock:
`:source path` will run any Git and gitsh commands in the file at `path`. It can modify the environment of the gitsh session it's called from, much like the `source` command in Unix shells.

### DIFF
--- a/lib/gitsh/commands/internal_command.rb
+++ b/lib/gitsh/commands/internal_command.rb
@@ -155,6 +155,37 @@ TXT
       end
     end
 
+    class Source < Base
+      USAGE_MESSAGE = 'usage: :source path'.freeze
+
+      def self.help_message
+        <<-TXT
+#{USAGE_MESSAGE}
+Runs the commands in the given file.
+TXT
+      end
+
+      def execute
+        if valid_arguments?
+          Gitsh::ScriptRunner.new(env: env).run(path)
+          true
+        else
+          env.puts_error USAGE_MESSAGE
+          false
+        end
+      end
+
+      private
+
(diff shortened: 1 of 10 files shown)


spec/integration/source_command_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+
+describe 'The :source command' do
+  context 'a source file is given' do
+    it 'executes the commands in the sourced file' do
+      GitshRunner.interactive do |gitsh|
```
jferris: These interactive tests are nice.

lib/gitsh/commands/internal_command.rb
```diff
@@ -155,6 +155,35 @@ def subject
...
+Runs the commands in the given file.
+TXT
+      end
+
+      def execute
+        if valid_arguments?
+          Gitsh::ScriptRunner.new(env: env).run(path)
+          true
+        else
+          env.puts_error 'usage: :source path'
```
jferris: Worth extracting this usage string into a method/constant?

jferris:
This looks good to me. I had one minor suggestion.

lib/gitsh/commands/internal_command.rb
georgebrock: Good idea.

spec/integration/source_command_spec.rb
georgebrock: Thanks!

georgebrock:
Thanks for the review, Joe. Mergin’

PR #218, opened 2015-01-19
Nested subshells

### LINKED ISSUE #198: Support nested subshells

georgebrock:
Update the parsing rules to support nested subshells.

Fixes #198

### DIFF
--- a/lib/gitsh/parser.rb
+++ b/lib/gitsh/parser.rb
@@ -115,7 +115,14 @@ module Gitsh
     end
 
     rule(:subshell) do
-      str('$(') >> (str(')').absent? >> any).repeat(0).as(:subshell) >> str(')')
+      str('$(') >> (subshell_content).as(:subshell) >> str(')')
+    end
+
+    rule(:subshell_content) do
+      (
+        (str('(') >> subshell_content >> str(')')) |
+        (str(')').absent? >> any)
+      ).repeat(0)
     end
 
     rule(:identifier) do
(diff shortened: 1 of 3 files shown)


lib/gitsh/parser.rb
```diff
@@ -115,7 +115,14 @@ def parse_and_transform(command)
...
 
     rule(:subshell) do
-      str('$(') >> (str(')').absent? >> any).repeat(0).as(:subshell) >> str(')')
+      str('$(') >> (subshell_content).as(:subshell) >> str(')')
+    end
+
+    rule(:subshell_content) do
+      (
+        (str('(') >> subshell_content >> str(')')) |
+        (str(')').absent? >> any)
```
gabebw: What's this do?

spec/integration/subshell_spec.rb
```diff
@@ -21,4 +21,14 @@
       expect(gitsh).to output /x in parent shell/
     end
   end
+
+  it 'supports nested subshells' do
+    GitshRunner.interactive do |gitsh|
```
gabebw: What a lovely abstraction.

JoelQ:
LGTM, although I'm not familiar with the syntax for the parser. :shipit:

lib/gitsh/parser.rb
georgebrock: This line is how you say “any single character except for `)`” in the Parslet DSL.

georgebrock:
Thanks for the reviews.

PR #219, opened 2015-01-20
Support for ~/.gitshrc

georgebrock:
The interactive runner will attempt to source a `~/.gitshrc` file.

### DIFF
--- a/lib/gitsh/interactive_runner.rb
+++ b/lib/gitsh/interactive_runner.rb
@@ -1,9 +1,11 @@
 require 'readline'
 require 'gitsh/completer'
+require 'gitsh/error'
 require 'gitsh/history'
 require 'gitsh/interpreter'
 require 'gitsh/prompter'
 require 'gitsh/readline_blank_filter'
+require 'gitsh/script_runner'
 require 'gitsh/term_info'
 
 module Gitsh
@@ -16,6 +18,7 @@ module Gitsh
       @history = opts.fetch(:history, History.new(@env, @readline))
       @interpreter = opts.fetch(:interpreter, Interpreter.new(@env))
       @term_info = opts.fetch(:term_info) { TermInfo.instance }
+      @script_runner = opts.fetch(:script_runner) { ScriptRunner.new(env: @env) }
     end
 
     def run
@@ -23,6 +26,7 @@ module Gitsh
       setup_readline
       handle_window_resize
       greet_user
+      load_gitshrc
       interactive_loop
     ensure
       history.save
@@ -30,7 +34,8 @@ module Gitsh
 
     private
 
-    attr_reader :history, :readline, :env, :interpreter, :term_info
+    attr_reader :history, :readline, :env, :interpreter, :term_info,
+      :script_runner
 
     def setup_readline
       readline.completion_append_character = nil
@@ -51,6 +56,15 @@ module Gitsh
 
     def greeting_enabled?
       env.fetch('gitsh.noGreeting') { 'false' } != 'true'
+    end
+
+    def load_gitshrc
+      script_runner.run(gitshrc_path)
+    rescue NoInputError
+    end
+
+    def gitshrc_path
+      "#{ENV['HOME']}/.gitshrc"
     end
 
     def interactive_loop
--- a/lib/gitsh/interactive_runner.rb
+++ b/lib/gitsh/interactive_runner.rb
@@ -15,8 +15,8 @@ module Gitsh
     def initialize(opts)
       @readline = opts.fetch(:readline) { ReadlineBlankFilter.new(Readline) }
       @env = opts[:env]
-      @history = opts.fetch(:history, History.new(@env, @readline))
-      @interpreter = opts.fetch(:interpreter, Interpreter.new(@env))
+      @history = opts.fetch(:history) { History.new(@env, @readline) }
+      @interpreter = opts.fetch(:interpreter) { Interpreter.new(@env) }
       @term_info = opts.fetch(:term_info) { TermInfo.instance }
       @script_runner = opts.fetch(:script_runner) { ScriptRunner.new(env: @env) }
     end
--- a/spec/integration/gitshrc_spec.rb
+++ b/spec/integration/gitshrc_spec.rb
@@ -1,0 +1,30 @@
+require 'spec_helper'
+
+describe 'A .gitshrc file in the home directory' do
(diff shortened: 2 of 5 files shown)


spec/support/file_system.rb
```diff
@@ -43,6 +37,23 @@ def chdir_and_allow_nesting(path)
       Dir.chdir(original_path)
     end
   end
+
+  private
+
+  def switch_home_directory(&block)
+    ENV['TEMP_HOME'] = 'TRUE'
+    orginal_home = ENV['HOME']
```
Euraldius: Should `orginal` be `original`?

Euraldius:
One small comment, otherwise LGTM

Also, it looks like `InteractiveRunner` is getting kinda hefty. It might require a decent refactor, but at some point it could be worth turning it into a Command object and synchronizing the interfaces of the objects it consumes

lib/gitsh/interactive_runner.rb
```diff
@@ -23,6 +26,7 @@ def run
       setup_readline
       handle_window_resize
       greet_user
+      load_gitshrc
```
teoljungberg: Don't we want to load the rc file before greeting the user? Might be good for i.e turning of the greeting message?

spec/support/file_system.rb
```diff
@@ -43,6 +37,23 @@ def chdir_and_allow_nesting(path)
       Dir.chdir(original_path)
     end
   end
+
+  private
+
+  def switch_home_directory(&block)
```
teoljungberg: How about considering using a library for this? https://github.com/defunkt/fakefs

lib/gitsh/interactive_runner.rb
georgebrock: I think I prefer it after the greeting: if there are commands in the rc file that produce output, it makes sense if that output appears after the greeting.

For example, here's my current `.gitshrc` file:

```
:set color.ui true

:echo
status
:echo
log --oneline -n 1
```

spec/support/file_system.rb
georgebrock: We need something more complete than stubbing for the integration specs.

Even if our Ruby code was hitting a fake file system, any Git commands that are called in the integration specs would still read the user's real `.gitconfig` file from their real home directory. Since the contents of the `.gitconfig` can make a big difference to the output of Git commands, it would be hard to write tests that make assertions about output if the output was different on different systems.

spec/support/file_system.rb
georgebrock: Yes, it should. Thanks.

lib/gitsh/interactive_runner.rb
```diff
@@ -16,6 +18,7 @@ def initialize(opts)
       @history = opts.fetch(:history, History.new(@env, @readline))
       @interpreter = opts.fetch(:interpreter, Interpreter.new(@env))
       @term_info = opts.fetch(:term_info) { TermInfo.instance }
+      @script_runner = opts.fetch(:script_runner) { ScriptRunner.new(env: @env) }
```
gabebw: Hm, why use a block here and for `@term_info` but not for `@interpreter` or `@history`?

spec/integration/gitshrc_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+          expect(gitsh).to output_no_errors
+        end
+      end
+    end
+  end
+
+  context 'when it does not exist' do
+    it 'does not cause an error' do
+      GitshRunner.interactive do |gitsh|
+        expect(File.exist?(gitshrc_path)).to be_false
```
gabebw: Does this need to be inside the `interactive` block? Does it make more sense to move it out?

spec/units/interactive_runner_spec.rb
```diff
@@ -25,6 +25,14 @@
       expect(readline).to have_received(:completion_proc=)
     end
 
+    it 'loads the ~/.gitshrc file' do
+      runner = build_interactive_runner
+      runner.run
```
gabebw: Should this have an empty line above it to separate it from the setup phase?

lib/gitsh/interactive_runner.rb
georgebrock: There should probably all use blocks, otherwise the defaults will be instantiated even when they're not used.

georgebrock: Fixed in ce563af98ca1d575f89ab7e8475403f9127dda45. Thanks, Gabe.

spec/integration/gitshrc_spec.rb
georgebrock: The `gitshrc_path` depends on the user's home directory, and one of the things that the `interactive` method does is switch to a fake home directory to avoid using a real `.gitconfig` file.

spec/units/interactive_runner_spec.rb
georgebrock: Good call. Fixed.

georgebrock:
Thanks for the reviews, everyone.

@gabebw @teoljungberg I didn't fix everything you commented on. Do my arguments seem sound, or does this need more work?

gabebw:
I am entirely convinced, and I award this PR my seal of approval.

georgebrock:
I gratefully receive your seal of approval, and shall merge forthwith.

teoljungberg:
I'm convinced aswell, go ahead and merge!

georgebrock:
Thanks, Teo. Mergin’

lib/gitsh/interactive_runner.rb
```diff
@@ -53,6 +58,16 @@ def greeting_enabled?
       env.fetch('gitsh.noGreeting') { 'false' } != 'true'
     end
 
+    def load_gitshrc
+      script_runner.run(gitshrc_path)
+    rescue NoInputError
+      false
```
mike-burns: This value is unused.

georgebrock: I was uncomfortable just leaving an empty rescue block. Do you think that—and an implicit `nil`—would be preferable to a potentially confusing unused value?

mike-burns: I'd find an empty `rescue` block more clear, actually: you're rescuing and discarding the `NoInputError`.

georgebrock: OK, I'm convinced. It's gone.

PR #221, opened 2015-01-21
Fix window size bugs

### LINKED ISSUE #144: Commands that are too long write over the command prompt
They should probably instead go to the next line.

screenshot from 2014-03-17 10 37 35

I'm using a Bash Shell on a Debian Machine.

georgebrock:
When gitsh receives a `SIGWINCH`, it informs Readline of the new terminal size.

Previously, the new terminal size was determined using the `tput cols` and `tput lines` commands. However, when Readline is sent the terminal size it sets `COLUMNS` and `LINES` environment variables, and if these environment variables are set `tput` will read them instead of actually checking the terminal size. This means that we were never actually passing the new terminal size to Readline, we were just passing whatever size we had when the environment variables were set, over and over again.

Determining the terminal size by trying each of the following in turn seems to work better:

1. `stty size` – this might fail, because some implementations of `stty` don't support `size`, e.g. on Solaris.
2. `env COLUMNS='' tput cols` and `env LINES='' tput lines` – this might fail, because some implementations of `tput` don't support `cols` or `lines` unless the environment variables are present.
3. `tput cols` and `tput lines` – this is what we have now.
4. Ignore the `SIGWINCH` entirely.

This also involved switching from using `Open3.capture3` to `IO.popen`, because the `stty` command only works if its input stream is a TTY, and `Open3.capture3` passes a pipe instead of `$stdin`. Unfortunately, this makes the testing more complex.

Fixes #144. Probably. I can no longer produce the [reproducible resize error](https://github.com/thoughtbot/gitsh/issues/144#issuecomment-69459773) from this branch, so it's promising.

### DIFF
--- a/lib/gitsh/interactive_runner.rb
+++ b/lib/gitsh/interactive_runner.rb
@@ -44,7 +44,10 @@ module Gitsh
 
     def handle_window_resize
       Signal.trap('WINCH') do
-        readline.set_screen_size(terminal.lines, terminal.cols)
+        begin
+          readline.set_screen_size(*terminal.size)
+        rescue Terminal::UnknownSizeError
+        end
       end
     end
 
--- a/lib/gitsh/interactive_runner.rb
+++ b/lib/gitsh/interactive_runner.rb
@@ -6,7 +6,7 @@ require 'gitsh/interpreter'
 require 'gitsh/prompter'
 require 'gitsh/readline_blank_filter'
 require 'gitsh/script_runner'
-require 'gitsh/term_info'
+require 'gitsh/terminal'
 
 module Gitsh
   class InteractiveRunner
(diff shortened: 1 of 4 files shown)


mike-burns:
That `IO.popen` sure is fun.

I've no complaints or concerns. The only thing that made me think was the fact that you rely on env(1) to set the environment variables (`COLUMNS`, `LINES`), but I couldn't find any quick solutions using `IO.popen`.

georgebrock:
I'm not too worried about depending on env(1): we're handling errors gracefully, and it's required by POSIX, as are the other dependencies I've used here (tput(1), stty(1) and `/dev/null`).

mike-burns:
Agreed. I have no further critique.

PR #224, opened 2015-01-24
Replace `mocha` with `rspec-mocks`

### LINKED ISSUE #215: Replace mocha with RSpec mocks
RSpec now has the mocking features that were missing a year ago and caused me to select mocha. Other developers seem to be surprised that the project isn't using RSpec mocks. Let's use mocha to make it easier for other developers to get involved.

teoljungberg:
- Replace `stub` with `double` from Rspec.
  - Use `spy` instead where applicable.
- Add `sequenced_double` helper to allow two arbitrary mock implementations to
  be called after each other.

Closes #215

### DIFF
--- a/spec/units/commands/git_command_spec.rb
+++ b/spec/units/commands/git_command_spec.rb
@@ -3,18 +3,21 @@ require 'gitsh/commands/git_command'
 
 describe Gitsh::Commands::GitCommand do
   let(:env) do
-    stub('Environment', {
+    double('Environment', {
       git_command: '/usr/bin/env git',
-      output_stream: stub(to_i: 1),
-      error_stream: stub(to_i: 2)
+      output_stream: double('OutputStream', to_i: 1),
+      error_stream: double('ErrorStream', to_i: 2)
     })
   end
 
-  before { Process.stubs(spawn: 1, wait: nil) }
+  before do
+    allow(Process).to receive(:spawn).and_return(1)
+    allow(Process).to receive(:wait)
+  end
 
   describe '#execute' do
     it 'spawns a process with the sub command and arguments' do
-      env.stubs(config_variables: {})
+      allow(env).to receive(:config_variables).and_return({})
       command = described_class.new(
         env,
         'commit',
@@ -33,10 +36,10 @@ describe Gitsh::Commands::GitCommand do
     end
 
     it 'passes on configuration variables from the environment' do
-      env.stubs(config_variables: {
+      allow(env).to receive(:config_variables).and_return(
         :'test.example' => 'This is an example',
         :'foo.bar' => '1'
-      })
+      )
       command = described_class.new(
         env,
         'commit',
--- a/spec/units/commands/internal/chdir_spec.rb
+++ b/spec/units/commands/internal/chdir_spec.rb
@@ -6,14 +6,14 @@ describe Gitsh::Commands::InternalCommand::Chdir do
 
   describe '#execute' do
     it 'returns true for correct directories' do
-      env = stub(:[]= => true, puts_error: true)
+      env = double('Environment', :[]= => true, puts_error: true)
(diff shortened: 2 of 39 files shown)


spec/support/sequence_double.rb
```diff
@@ -0,0 +1,12 @@
+# Sequence two arbitrary mock implementations after each other. For example,
+# one can use this to sequence a return after an exception.
+#
+#   helloer = sequenced_double(double("Hello"), -> { raise "yo" }, lambda { 1 })
+#   SayHello.run(helloer)
+def sequenced_double(dbl, *actions)
```
georgebrock: This method feels a little too clever to me. The `__send__` suggests we're using a private interface which we definitely shouldn't be doing.

Since this is only used in a couple of places, something like this might be simpler:

```ruby
method_results = [-> { raise Interrupt }, -> { nil }].to_enum
allow(some_object).to receive(:some_method) { method_results.next.call }
```

If this is hard to read, we could replace the `method_results` array with a custom object that provides a more readable interface, e.g.:

```ruby
method_results = StubbedMethodResult.raise(Interrupt).then_return(nil)
allow(some_object).to receive(:some_method) { method_results.next_result }
# or, if we want to get crazy with #to_proc methods
allow(some_object).to receive(:some_method, &method_results)
```

spec/units/commands/git_command_spec.rb
```diff
@@ -33,7 +36,7 @@
     end
 
     it 'passes on configuration variables from the environment' do
-      env.stubs(config_variables: {
+      allow(env).to receive(:config_variables).and_return({
```
georgebrock: The braces aren't needed here.

spec/units/commands/internal/chdir_spec.rb
```diff
@@ -6,14 +6,14 @@
 
   describe '#execute' do
     it 'returns true for correct directories' do
-      env = stub(:[]= => true, puts_error: true)
+      env = double("Environment", :[]= => true, puts_error: true)
```
georgebrock: This project prefers single quotes (it was started before we changed the style guide to recommend double quotes).

spec/units/commands/internal/chdir_spec.rb
```diff
@@ -6,14 +6,14 @@
...
       command = described_class.new(env, 'cd', arguments('./'))
 
       expect(command.execute).to be_truthy
     end
 
     it 'returns false with invalid arguments' do
-      env = stub(:[]= => true, puts_error: true)
+      env = double("Environment", :[]= => true, puts_error: true)
```
georgebrock: Double quotes.

georgebrock:
Thanks for this @teoljungberg. I left a few minor comments about style, and I think the `#sequenced_double` method needs some work, but it's looking good.

spec/support/sequence_double.rb
teoljungberg: I agree, `sequence_double` is a big 'ol hack, it definately needs some love. I like the enumerable solution, I'll give it a whirl and try how it feels

spec/units/commands/internal/chdir_spec.rb
teoljungberg: I know, I recently started using the habbit of using double-quotes myself. But it seems hard getting that out of my muscle memory. Sorry!

georgebrock: No problem, I do it too. Just trying to keep the code as consistent as possible.

teoljungberg:
@georgebrock comments addressed, fixed and pushed

georgebrock:
@teoljungberg Looking good! Looks like there's a legitimate spec failure, then this is ready to merge.

I guess it should probably be squashed down; I don't think keeping the separate commits will make the history more clear.

teoljungberg:
Thanks for the feedback @georgebrock, merging

Issue #134, opened 2014-03-10
Execute shell commands through a shell


georgebrock:
Not using a shell leads to unexpected behaviour, e.g. without a shell to interpret the `~` in the argument, the command `!vim ~/foo` will edit the file `./~/foo`

georgebrock:
Work in progress on the `shell-commands-through-shell` branch.

PR #225, opened 2015-02-20
Refactor shell commands

georgebrock:
Refactors `Gitsh::Commands::ShellCommand` and `Gitsh::Commands::GitCommand`

* Extract a `Gitsh::ShellCommandRunner` from `Gitsh::Commands::ShellCommand`.
* Replace inheritance with composition.
* Pave the way to solve #134 without modifying the behaviour of Git commands.

### DIFF
--- a/spec/units/commands/git_command_spec.rb
+++ b/spec/units/commands/git_command_spec.rb
@@ -44,6 +44,7 @@ describe Gitsh::Commands::GitCommand do
         env,
         'commit',
         arguments('-m', 'A test commit'),
+        shell_command_runner: mock_runner,
       )
 
       command.execute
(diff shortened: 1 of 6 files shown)


georgebrock:
(Spec failure is unrelated, and will be fixed by a future C extension)

spec/units/commands/git_command_spec.rb
```diff
@@ -2,43 +2,34 @@
...
+      command = described_class.new(env, "commit", arguments("-m", "Some stuff"))
+      result = command.execute
 
-      expect(Process).to have_received(:spawn).with(
-        '/usr/bin/env', 'git',
-        'commit',
-        '-m', 'A test commit',
-        out: env.output_stream.to_i,
-        err: env.error_stream.to_i
+      expect(Gitsh::ShellCommandRunner).to have_received(:new).with(
```
mike-burns: The strategy in this project is to pass all classes as arguments to the constructor, and to pass a double in the test. Here, you do it differently. Any reason?

(IIRC, `Process.spawn` was mocked here because it was tricky to do it otherwise.)

georgebrock: No reason, so I've updated the code to be consistent. Thanks.

PR #234, opened 2015-08-07
Support ~ in paths given to `:source`

georgebrock:
The `:cd` command already supported `~` in paths, but `:source` did not. This commit fixes that inconsistency.

### DIFF
--- a/spec/integration/source_command_spec.rb
+++ b/spec/integration/source_command_spec.rb
@@ -13,6 +13,18 @@ describe 'The :source command' do
         expect(gitsh).to output /Yes it did!/
       end
     end
+
+    it 'expands ~ in paths' do
+      GitshRunner.interactive do |gitsh|
+        write_file("#{ENV['HOME']}/.gitshrc", ':set source_worked "True"')
+
+        gitsh.type ':source ~/.gitshrc'
+        gitsh.type ':echo $source_worked'
+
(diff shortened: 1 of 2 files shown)


spec/integration/source_command_spec.rb
```diff
@@ -13,6 +13,18 @@
         expect(gitsh).to output /Yes it did!/
       end
     end
+
+    it 'expands ~ in paths' do
+      GitshRunner.interactive do |gitsh|
+        write_file("#{ENV['HOME']}/.gitshrc", ':set source_worked "True"')
```
teoljungberg: Should `write_file` also be able to take a `~` as `$HOME`?

mike-burns:
Yeah, LGTM.

(I went looking for a way to make sure that this applies to every object that can produce a path, including `#path` methods you'll add in the future, and I concluded: no idea, let's tackle that when it happens.)

spec/integration/source_command_spec.rb
georgebrock: Maybe, but I quite like how explicit this is: it means the test makes it clear what `~` should expand to.

teoljungberg: In that case - add it when you feel the pain

georgebrock:
Thanks for the reviews. Merging.

PR #235, opened 2015-08-07
Use non-zero exit status on fatal errors.

### LINKED ISSUE #233: Incorrect exit status when gitsh crashes
When gitsh encounters a fatal error and crashes, it should exit with a non-zero status.

This looks like it'll be very easy to solve: we're just not exiting when we handle an error https://github.com/thoughtbot/gitsh/blob/master/src/gitsh.rb.in#L11

(Originally reported by @bernerdschaefer as part of #230)

georgebrock:
In which I correct an embarrassing oversight.

Fixes #233

### DIFF
--- a/src/gitsh.rb.in
+++ b/src/gitsh.rb.in
@@ -9,4 +9,5 @@ begin
   Gitsh::CLI.new.run
 rescue => e
   $stderr.puts "gitsh: Error: #{e.message}"
+  exit 1
 end


teoljungberg:
Do we need any tests for this?

georgebrock:
@teoljungberg We haven't had any so far. Maybe this problem existing for so long indicates that we should have! Any thoughts on how we should test this, given that this error handling code is dsigned to catch unexpected bugs?

PR #237, opened 2015-09-03
Raise UnsetVariableError for magic variables.

### LINKED ISSUE #230: gitsh crashes when referencing $_prior does not exist
Here's the sequence of commands I performed to encounter the error:

```
> gitsh
checkout -b feature-branch
checkout master
branch -d $_prior
branch -d $_prior
```

And here's a simplified case:

```
> gitsh
!echo $_prior
```

### What happened

gitsh printed the following message and exited:

```
gitsh: Error: no implicit conversion of nil into String
```

### What I expected to happen

I expected gitsh to do one of the following, and continue accepting commands:

- print error about the variable not existing
- run the command with an empty value, and let errors bubble up

georgebrock:
The magic variables that gitsh provides all depend on certain state, e.g. the `$_merge_base` variable is only applicable during a merge.

This commit brings magic variables into line with regular user-defined variables, but raising an `UnsetVariableError` if they aren't currently set, i.e. if the state from which the variable derives its value doesn't exist.

Fixes #230

### DIFF
--- a/lib/gitsh/magic_variables.rb
+++ b/lib/gitsh/magic_variables.rb
@@ -21,15 +21,22 @@ module Gitsh
     end
 
     def _prior
-      repo.revision_name('@{-1}')
+      repo.revision_name('@{-1}') ||
+        raise(UnsetVariableError, 'No prior branch')
     end
 
     def _merge_base
-      repo.merge_base('HEAD', 'MERGE_HEAD')
+      repo.merge_base('HEAD', 'MERGE_HEAD').tap do |merge_base|
+        if merge_base.empty?
+          raise UnsetVariableError, 'No merge in progress'
+        end
+      end
     end
 
     def _rebase_base
-      read_file(['rebase-apply', 'onto']) || read_file(['rebase-merge', 'onto'])
+      read_file(['rebase-apply', 'onto']) ||
+        read_file(['rebase-merge', 'onto']) ||
+        raise(UnsetVariableError, 'No rebase in progress')
     end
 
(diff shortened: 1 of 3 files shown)


teoljungberg:
One testing-question/recomendation. Besides that, merge at your descretion

spec/units/magic_variables_spec.rb
```diff
@@ -22,6 +22,15 @@
         expect(magic_variables.fetch(:_prior)).to eq 'a-branch-name'
         expect(repo).to have_received(:revision_name).with('@{-1}')
       end
+
+      it 'raises when there is no prior branch' do
+        repo = double('GitRepository', revision_name: nil)
+        magic_variables = described_class.new(repo)
+
+        expect { magic_variables.fetch(:_prior) }.to raise_exception(
+          Gitsh::UnsetVariableError,
```
georgebrock: We're asserting that sensible errors are output in the integration tests. I'm not sure there's much value in repeating them here.

lib/gitsh/magic_variables.rb
```diff
@@ -21,15 +21,25 @@ def available_variables
     end
 
     def _prior
-      repo.revision_name('@{-1}')
+      repo.revision_name('@{-1}').tap do |revision|
+        if revision.nil?
+          raise UnsetVariableError, 'No prior branch'
+        end
```
georgebrock: Good call. Done.

georgebrock:
Thanks for the reviews.

PR #238, opened 2015-09-03
More correct vendorize script.

### LINKED ISSUE #203: Vendorize script assumes `lib` should be required
The `vendorize` script assumes that the correct load path for any given gem is:

```ruby
"../#{gem.name}-#{gem.version}/lib"
```

It would be more correct to support multiple paths per gem, doing something like this:

```ruby
gem.require_paths.map do |path|
  "../#{gem.name}-#{gem.version}/#{path}"
end
```

Currently, for all of the gem dependencies of gitsh, `require_paths == ["lib"]`, so this isn't causing any problems in practice. If we add more gem dependencies in the future it could cause problems.

georgebrock:
The vendorize script assumed that the correct require path for any given gem was "lib", and only "lib". While this assumption holds true for all of the gems we currently depend on, removing it will make it easier to add future dependencies without potentially obscure bugs.

Fixes #203

### DIFF
--- a/vendor/vendorize
+++ b/vendor/vendorize
@@ -22,9 +22,11 @@ File.open(SETUP_PATH, 'w') do |setup_file|
         gem.name
       )
 
-      setup_file << "$LOAD_PATH.unshift(File.expand_path(%s, __FILE__))\n" % [
-        "../#{gem.name}-#{gem.version}/lib".inspect
-      ]
+      template = "$LOAD_PATH.unshift(File.expand_path(%s, __FILE__))"
+      gem.require_paths.each do |path|
+        full_path = "../#{gem.name}-#{gem.version}/#{path}"
+        setup_file.puts(template % full_path.inspect)
+      end
     end
   end
 end


vendor/vendorize
```diff
@@ -22,9 +22,12 @@ File.open(SETUP_PATH, 'w') do |setup_file|
         gem.name
       )
 
-      setup_file << "$LOAD_PATH.unshift(File.expand_path(%s, __FILE__))\n" % [
-        "../#{gem.name}-#{gem.version}/lib".inspect
-      ]
+      template = "$LOAD_PATH.unshift(File.expand_path(%s, __FILE__))\n"
+      gem.require_paths.each do |path|
+        setup_file << template % [
```
mike-burns: If you do `setup_file.puts` instead of `setup_file <<`, would that get rid of the `\n` in the template?

mike-burns:
Good idea.

vendor/vendorize
georgebrock: Good call, done.

mike-burns:
LGTM.

PR #241, opened 2015-10-11
Update installation instructions for newest release

jjb:

PR #245, opened 2015-12-02
Update INSTALL for 0.10

adrianheine:


mike-burns:
@georgebrock it looks like `INSTALL.in` generates `INSTALL`, but it didn't for 0.10.

georgebrock:
Thanks for this! #241 makes the same changes, and was opened slightly earlier, so I'll merge that version. I've also opened issue #247 to add this to the release make target, so we won't miss it next time.

PR #246, opened 2016-02-24
Add link to video on Upcase

christoomey:
We've made the intro video on Upcase free to all users (no subscription required), so I'd like to add a link to the video from the README.

Happy to revise the text if something else makes more sense.


georgebrock:
I like it!

I suggest we drop the word “free”. Most videos on the Web are free, and outside the context of the Upcase Web site the distinction doesn't make sense.

christoomey:
@georgebrock makes sense. I've updated to remove "free". Look good?

PR #249, opened 2016-05-13
Execute shell commands through /bin/sh

### LINKED ISSUE #134: Execute shell commands through a shell

georgebrock:
Resolves #134

### DIFF
--- a/lib/gitsh/commands/shell_command.rb
+++ b/lib/gitsh/commands/shell_command.rb
@@ -3,6 +3,10 @@ require 'gitsh/shell_command_runner'
 module Gitsh
   module Commands
     class ShellCommand
+      SHELLWORDS_WHITELIST = 'A-Za-z0-9_\-.,:\/@\n'.freeze
+      GLOB_WHITELIST = '\*\[\]!\?\\\\'.freeze
+      SHELL_CHARACTER_FILTER = /([^#{SHELLWORDS_WHITELIST}#{GLOB_WHITELIST}])/
+
       def initialize(env, command, args, options = {})
         @env = env
         @command = command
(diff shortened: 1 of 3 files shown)


lib/gitsh/commands/shell_command.rb
```diff
@@ -22,11 +22,15 @@ def execute
...
+        [
+          '/bin/sh',
+          '-c',
+          [command, arg_values].flatten.join(' '),
+        ]
       end
 
       def arg_values
-        args.values(env)
+        args.values(env).map { |arg| arg.gsub(' ', '\ ') }
```
mike-burns: Should we use `Shellwords` here?

georgebrock: Good call.

mike-burns:
After `Shellwords`: LGTM.

georgebrock:
I'm having a little trouble with `Shellwords`: some things (like spaces in the middle of arguments) ought to be escaped, but other things (like globs) ought not be escaped. `Shellwords` will escape all of them.

A few examples:

* `!echo Hello world` should execute `/bin/sh -c "echo Hello world"`
* `!echo "Hello world"` should execute `/bin/sh -c "echo Hello\ world"`
* `!ls *` should execute `/bin/sh -c "ls *"`

mike-burns:
I don't see a solution using `Shellwords` directly.

Interestingly, `shellwords.rb` is incredibly short and mostly consists of:

```ruby
def shellescape(str)
  str = str.to_s
  return "''" if str.empty?
  str = str.dup
  str.gsub!(/([^A-Za-z0-9_\-.,:\/@\n])/, "\\\\\\1")
  str.gsub!(/\n/, "'\n'")
  str
end
```

(Read the file itself for more details.)

Seems smart to use a whitelist like they do instead of a blacklist. Maybe this regexp: `[^A-Za-z0-9_\-.,:\/@\n*?\[\!]` -- pulling from [glob(7)](http://man.openbsd.org/glob.7).

georgebrock:
@mike-burns All great ideas! I added custom escaping, and tried to make it explicit what was based on `Shellwords` and what was added to avoid escaping globbing patterns. Let me know what you think.

lib/gitsh/commands/shell_command.rb
```diff
@@ -1,8 +1,14 @@
+require 'shellwords'
```
mike-burns: Is this used?

mike-burns:
Looks great!

PR #251, opened 2016-05-20
Use bundle exec rspec to run specs

calleerlandsson:
On Mac OS X 10.11.4 and Debian 8 using system Ruby the `make check` task
would fail on the `units/terminal_spec.rb` file with a `NoMethodError`
for `allow` on `CommandStubber`. This was probably because rspec-mocks
did not properly laod.

Running specs using `bundle exec rspec` solves this.


georgebrock:
LGTM. Thanks!

calleerlandsson:
Thanks @georgebrock. I'll merge this.

PR #252, opened 2016-05-20
Ignore sequential duplicate history entries

### LINKED ISSUE #68: Erase duplicate sequential commands from history.
If the same command has been issued several times in a row, I have to up-arrow an equal number of times to get past it in the history. Would be nice if this could be compacted, like bash does with HISTCONTROL=ignoredups

calleerlandsson:
Previously when the same command had been issued several times in a row,
users would have to press the up arrow an equal number of times to get
past it in the history. Some users found this tiresome.

This commit changes the Readline filter to filter out sequential
duplicate history entires in addition to blank ones.

This fixes #68.

This PR also includes the changes from #251 but will be rebased on master before merging.


lib/gitsh/readline_history_filter.rb
```diff
@@ -0,0 +1,21 @@
+require 'gitsh/module_delegator'
+
+class ReadlineHistoryFilter < ModuleDelegator
+  def readline(prompt, add_hist = false)
+    module_delegator_target.readline(prompt, add_hist).tap do |input|
+      if add_hist && input && (input.strip.empty? || duplicate?(input))
```
georgebrock: There's a lot going on in this line. Could we extract a predicate method? `#should_not_have_added_to_history?` or something?

lib/gitsh/readline_history_filter.rb
```diff
@@ -0,0 +1,21 @@
...
+        history.pop
+      end
+    end
+  end
+
+  private
+
+  def duplicate?(input)
+    history.length > 1 && input == history[-2]
```
georgebrock: Should be use `history[-1] == history[-2]` here, instead of passing `input`?

spec/units/readline_history_filter_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      fake_readline.type('hello')
+      readline_history_filter.readline('>', true)
+      fake_readline.type('hello')
+      readline_history_filter.readline('>', true)
+
+      expect(readline_history_filter::HISTORY.to_a).to eq(['hello'])
+    end
+
+    it 'does not delete non-sequential duplicate history' do
```
georgebrock: How about `it 'enters non-sequential duplicate history'` here? Seems more consistent with the other descriptions.

lib/gitsh/readline_history_filter.rb
georgebrock: To be more clear: it took me a couple of readings to work out that `input` should be the same as `history[-1]` and that's why we were using `history[-2]` to look for duplicates. At first glance it looked like we were skipping over the most recent history entry.

lib/gitsh/readline_history_filter.rb
georgebrock: I guess we'd still need `add_hist && should_not_have_added_to_history?`, but that still seems more clear.

georgebrock:
Left a bunch of minor comments, but overall this looks great. Thanks, Calle!

lib/gitsh/readline_history_filter.rb
calleerlandsson: Great call, done in 98ab814.

lib/gitsh/readline_history_filter.rb
calleerlandsson: Good idea! Changed in 98ab814.

spec/units/readline_history_filter_spec.rb
calleerlandsson: Much better. Changed in 98ab814.

calleerlandsson:
Thanks for reviewing @georgebrock! Do you want to look it through one last time before I merge?

georgebrock:
Looks good to me!

calleerlandsson:
Great! I'm squashing and merging.

PR #253, opened 2016-05-27
Provide a helpful message if git is not executable

### LINKED ISSUE #193: gitsh with a missing git is confusing
```
% env PATH= gitsh
env: gitsh: Permission denied
```

It should instead say that it cannot find a `git` program, that you should pass `--git` or configure `gitsh.gitCommand`, and to see gitsh(1) for more details.

calleerlandsson:
Display a helpful message and exit early when gitsh is installed on a
system without git or when a non-existent git is specified using the
`--git` option.

This fixes #193.

### DIFF
--- a/lib/gitsh/cli.rb
+++ b/lib/gitsh/cli.rb
@@ -24,7 +24,9 @@ module Gitsh
       parse_arguments
       if unparsed_args.any?
         exit_with_usage_message
-      elsif script_file
+      end
+      ensure_executable_git
+      if script_file
         run_script
       else
         interactive_runner.run
@@ -81,6 +83,22 @@ module Gitsh
           exit EX_OK
         end
       end
+    end
+
+    def ensure_executable_git
+      IO.popen(env.git_command).close
+    rescue Errno::ENOENT
+      env.puts_error(
+        "gitsh: #{env.git_command}: No such file or directory\nEnsure git is "\
+        'on your PATH, or specify the path to git using the --git option',
+      )
+      exit EX_UNAVAILABLE
+    rescue Errno::EACCES
+      env.puts_error(
+        "gitsh: #{env.git_command}: Permission denied\nEnsure git is "\
+        'executable',
+      )
+      exit EX_UNAVAILABLE
     end
   end
 end
--- a/spec/integration/running_scripts_spec.rb
+++ b/spec/integration/running_scripts_spec.rb
@@ -6,7 +6,7 @@ describe 'Executing gitsh scripts' do
       in_a_temporary_directory do
         write_file('myscript.gitsh', "init\n\ncommit")
 
-        expect("#{gitsh} --git #{fake_git} myscript.gitsh").to execute.
+        expect("#{gitsh} --git #{fake_git_path} myscript.gitsh").to execute.
           successfully.
           with_output_matching(/^Fake git: init\nFake git: commit\n$/)
       end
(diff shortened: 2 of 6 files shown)


lib/gitsh/cli.rb
```diff
@@ -82,5 +84,15 @@ def option_parser
         end
       end
     end
+
+    def ensure_executable_git
+      IO.popen(env.git_command).close
```
teoljungberg: Does this impact the performance in any way?

teoljungberg:
I rebuilt CI in hopes that it will fix the segfault.

The code looks good to me, I had a minor question.

lib/gitsh/cli.rb
calleerlandsson: It doesn't seem like it. `time bin/gitsh test` where test is an empty file takes roughly 0.25s on my system.

lib/gitsh/cli.rb
```diff
@@ -82,5 +84,15 @@ def option_parser
...
+    rescue Errno::ENOENT
+      env.puts_error(
+        "gitsh: No such executable git: #{env.git_command}. Please " \
+        'specify an executable git using the --git option.',
+      )
+      exit EX_DATAERR
```
georgebrock: Maybe `EX_UNAVAILABLE` (69) would be more appropriate here?

> A service is unavailable.  This can occur if a support program or file does not exist.

lib/gitsh/cli.rb
```diff
@@ -82,5 +84,15 @@ def option_parser
         end
       end
     end
+
+    def ensure_executable_git
+      IO.popen(env.git_command).close
+    rescue Errno::ENOENT
```
georgebrock: We should possibly also rescue from `Errno::EACCES` here, with a different error message indicating the supposed Git binary wasn't executable.

lib/gitsh/cli.rb
```diff
@@ -82,5 +84,15 @@ def option_parser
...
+      env.puts_error(
+        "gitsh: No such executable git: #{env.git_command}. Please " \
+        'specify an executable git using the --git option.',
```
georgebrock: Since this is specifically for `ENOENT`, I wonder if we should drop the mention of “executable”? How about:

    gitsh: #{env.git_command}: No such file or directory
    Ensure git is on your PATH, or specify the path to git using the --git option

The first line of this would be exactly what `ls`, `cat`, etc. would output, which seems like a nice consistency to aim for.

spec/integration/running_scripts_spec.rb
```diff
@@ -36,8 +36,4 @@
   def gitsh
     File.expand_path('../../../bin/gitsh', __FILE__)
   end
-
-  def fake_git
-    File.expand_path('../../../spec/fixtures/fake_git', __FILE__)
-  end
```
georgebrock: Thanks for cleaning up the duplication here.

georgebrock:
Left a few minor comments, but overall this LGTM!

lib/gitsh/cli.rb
calleerlandsson: Good call! Done in 8a4246ee3077672948eca4a12b71359ee3657b68.

lib/gitsh/cli.rb
calleerlandsson: Great idea! Done in 1b721c5.

lib/gitsh/cli.rb
calleerlandsson: Good catch. Done in e68c2d8.

calleerlandsson:
Thanks for reviewing @georgebrock! I've updated the PR according to your comments.

georgebrock:
Looks great. Thanks, Calle.

calleerlandsson:
Sorry for dropping the ball on this one! I'll rebase, squash, and merge.

georgebrock:
Thanks!

Issue #255, opened 2016-06-15
Use custom Readline integration


georgebrock:
Various outstanding issues are impossible to solve without access to GNU Readline features that aren't exposed by the default Ruby bindings (#114, #150, #189). Alternative solutions have been tried, but have failed (#49, #121).

We need to bite the bullet and replace Ruby's Readline module with our own C extension.

This is work-in-progress on the `gb-custom-readline-extension` branch.

* [x] Copy Ruby's `Readline` module to the gitsh repository
* [x] Rename the module to `Gitsh::LineEditor`
* [x] Add the custom C extension to the gitsh autotools build configuration
* [x] Port Ruby's Readline tests to the gitsh repository
* [x] Support Ruby versions earlier than 2.1 (they [fail because the C interface has changed][failures])
* [x] Fix CI for Ruby 2.0.0 (this works for me locally, but not on CI—might be a patch-level version difference)
* [x] Ensure `extconf.rb` finds the right Readline implementation (by default it'll only look in places defined in `RbConfig::CONFIG`, which may or may not be useful depending on how and when Ruby was built)
* [x] Either explicitly support libedit (make sure the test suite passes with it) or explicitly drop it (make sure compilation fails with it)
* [x] Update Homebrew setup to require Readline
* [x] Update documentation
    * `CONTRIBUTING.md`: there are extra steps that a developer might need to take
    * `INSTALL.in`: the readline/libedit section is out of date
* [x] Refactor ported tests (there's a lot of duplication, and some of them are pretty hard to read)
* [x] Expand `Gitsh::LineEditor` specs for `other character settings`
* [ ] Expose Readline's `rl_char_is_quoted_p` function (see [georgebrock/readline-extensions])

[failures]: https://travis-ci.org/thoughtbot/gitsh/jobs/137798497
[georgebrock/readline-extensions]: https://github.com/georgebrock/readline-extensions

georgebrock:
Opened a PR (#256) to get feedback on this, even though it's still work in progress.

Issue #189, opened 2014-10-13
Tab completion doesn't work for arguments containing spaces


georgebrock:
A solution is in progress on the `gb-custom-readline-extension` branch, after which the `gb-tab-complete-spaces` branch may become relevant again.

See discussion on #121 for details.

georgebrock:
We're now using a custom GNU Readline integration, which means we should be able to fix this by:

* Making sure spaces in completions are escaped with a `\`,
* exposing GNU Readline's `rl_char_is_quoted_p` setting (see https://github.com/georgebrock/readline-extensions for an example), and
* setting a `char_is_quoted` proc that recognises a `\` followed by a space as a quoted space.

The same approach will probably fix #150.

georgebrock:
This is work-in-progress on the `gb-tab-complete-spaces` branch.

* [x] Support escaped literals in unquoted strings in `Gitsh::Parser`
* [x] Quote output from the `Gitsh::Completer` (e.g. with `some file.txt` typing <kbd>add some<kbd>tab</kbd></kbd> results in `add some\ file.txt` instead of `add some file.txt`)
* [x] Don't quote output from the `Gitsh::Completer` when completing a quoted string (e.g. with `some file.txt` typing <kbd>add "some<kbd>tab</kbd></kbd> results in `add "some file.txt"` instead of `add "some\ file.txt"`)
* [x] Expose Readline's `rl_char_is_quoted_p` function
* [x] Write a quoting detection proc to handle `\` escaped characters (so that typing <kbd>add some\ f<kbd>tab</kbd></kbd> results in `"some\ f"` being passed to `Gitsh::Completer` instead of just `"f"`)
* [x] Handle escaped escape characters in the quoting detection proc (e.g. the space in `foo\\ bar` is not escaped, but the space in `foo\\\ bar` is)
* [ ] Unquote input to the `Gitsh::Completer` so it can match against the options (e.g. when `"some\ f"` is passed to `Gitsh::Completer`, look for options matching `"some f"`)
* [ ] Make sure the `Gitsh::Completer` escaping correctly handles completing an unquoted argument after a quoted argument (e.g. <kbd>add "foo" some<kbd>tab</kbd></kbd> should escape the completions for `"some"`; using `rl_completion_found_quote` and `rl_completion_quote_character` instead of the current naive quote detection will probably help here)
[...]

Issue #150, opened 2014-03-27
Tab completion doesn't work for file names containing parentheses


neilkimmett:
Say I type in `diff MyCoolApp.xcodepr` and hit tab, I would expect it to be tab completed to `diff MyCoolApp.xcodeproj/` but it isn't :disappointed:

georgebrock:
We're now using a custom GNU Readline integration, which means we should be able to fix this by:

* Making sure `(` and `)` in completions are escaped to `\(` and `\)`,
* exposing GNU Readline's `rl_char_is_quoted_p` setting (see https://github.com/georgebrock/readline-extensions for an example), and
* setting a `char_is_quoted` proc that recognises `\(` as a quoted `(`, etc.

The same approach will probably fix #189.

neilkimmett:
Thats awesome news, thanks for following up on this @georgebrock. I didn't have this problem for a while but now that I work at Harry's I'm seeing the same issues with the apostrophe character in filenames. Let me know when theres something to play with and I'll be happy to help test it out.

georgebrock:
Fixed by #259.

PR #256, opened 2016-06-23
Custom readline extension

georgebrock:
This is still work in progress, but it's got pretty big and pretty complex.

Details of what this is, what's done, and what's still to do are on issue #255.


Makefile.am
```diff
@@ -1,5 +1,5 @@
 AUTOMAKE_OPTIONS = foreign
-SUBDIRS = src man lib/gitsh vendor spec
+SUBDIRS = src man vendor spec ext/gitsh lib/gitsh
```
mike-burns: Does this order matter? Or: was this a technical change that I should learn about?

ext/gitsh/src/line_editor.c
```diff
@@ -0,0 +1,1973 @@
+/************************************************
```
mike-burns: Does this file differ from upstream? Is it possible to clarify which is copied and which isn't? Or perhaps use e.g. `#include` to separate the files?

spec/units/line_editor_blank_filter_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'spec_helper'
```
mike-burns: Great pattern you've set up in this file -- easy to read.

spec/units/line_editor_spec.rb
```diff
@@ -0,0 +1,702 @@
+require 'spec_helper'
```
mike-burns: This file is exciting. Not sure I understand all of it on a first read. Anything in particular I should focus on?

ext/gitsh/src/line_editor.c
georgebrock: ef19570 introduced an exact copy of the upstream module, and subsequent commits made changes to it.

You can see just the changes here: https://github.com/thoughtbot/gitsh/compare/ef19570...gb-custom-readline-extension#diff-9

Makefile.am
georgebrock: I don't believe this matters. Not sure why I changed it now (I did the initial setup for this branch more than a year ago)

spec/units/line_editor_blank_filter_spec.rb
georgebrock: Credit where it's due: I just renamed this file, it was @sgrif and @adam89 who first wrote it.

spec/units/line_editor_spec.rb
georgebrock: It's mostly a port of the Ruby test suite's Readline tests, converted from minitest to RSpec, and better organised by method.

I intend to do more refactoring here before this is merged.

ext/gitsh/src/line_editor.c
```diff
@@ -0,0 +1,1973 @@
...
+readline_s_set_special_prefixes(VALUE self, VALUE str)
+{
+    if (!NIL_P(str)) {
+        OutputStringValue(str);
+        str = rb_str_dup_frozen(str);
+#if defined(HAVE_RB_OBJ_HIDE)
+        rb_obj_hide(str);
+#else
+        RBASIC(str)->klass = 0;
+#endif
```
mike-burns: Maybe this could do the same preprocessor trick as `readline_s_delete_bytes`:

```c
#if !defined(HAVE_RB_OBJ_HIDE)
#define rb_obj_hide(str) do{ RBASIC(str)->klass = 0; } while(0);
#end
```

(Same with `rb_obj_reveal` below.)

Then you won't need to modify these functions directly.

georgebrock: Could you explain what the `do{ … } while(0);` is for? I've not seen that before.

mike-burns: In the exact case that you showed, it will work fine without the `do/while`. However, C macros break in cases like this:

```c
#define DO_THE_THING(x) thing1(x); thing2(x);
if (true)
        DO_THE_THING(str);
else
        warnx("no thing found");
```

So the `do/while` was a habit and not technically needed in this case: this case is where the macro expands to exactly one statement, and it is not used inside a branch of a braceless block.

georgebrock: Makes sense, thanks.

georgebrock:
@mike-burns Thanks for the feedback so far. I think this is now ready to go.

Since you last looked, I've:

* Dropped support for libedit,
* dropped support for a bunch of less useful readline features,
* refactored the `Gitsh::LineEditor` specs so that they're somewhat readable,
* used macros in C code in the ways you suggested, and
* updated the documentation.

This PR appears to solve #114. It also sets the stage for solving #150 and #189, but doesn't take  the crucial last step of exposing the `rl_char_is_quoted_p` setting; there's already enough going on here without adding new functionality, so I'll save that for another PR.

CONTRIBUTING.md
```diff
@@ -13,15 +13,33 @@ Here's a quick guide to begin contributing:
 
         git clone https://github.com/thoughtbot/gitsh.git
 
+2. Ensure GNU Readline is installed, e.g. on OS X you would run:
+
+        brew install readline
+
 2. Build the generated files. Some Ruby files in `gitsh` are generated by the
```
mike-burns: Want to update this `2` to a `3`, and so on?

ext/gitsh/src/line_editor.c
```diff
@@ -0,0 +1,1764 @@
+/************************************************
+
+  Based on Ruby 2.3.0's
```
mike-burns: :+1:

lib/gitsh/line_editor.rb
```diff
@@ -0,0 +1,31 @@
...
+        len = args[1]
+      else
+        raise ArgumentError,
+          "wrong number of arguments (given #{args.length}, expected 0..2)"
+      end
+
+      byte_beg = line_buffer[0...beg].bytesize
+      byte_len = line_buffer[beg, len].bytesize
+
+      self.delete_bytes(byte_beg, byte_len)
```
mike-burns: :tada:

mike-burns:
Excellent!

Do you know: does this behave well on a Ruby that was built against libedit? My understanding is that it will work fine: you're linking against Readline and ignoring whatever Ruby does.

LGTM.

CONTRIBUTING.md
georgebrock: Good catch.

ext/gitsh/src/line_editor.c
```diff
@@ -0,0 +1,1764 @@
...
+
+#if !defined(HAVE_RB_OBJ_REVEAL)
+#define rb_obj_reveal(str, newKlass) do {\
+  RBASIC(str)->klass = newKlass;\
+} while (0);
+#endif
+
+
+/*
+ * Document-class: Readline
```
georgebrock: I should probably change the documentation in this file: it refers to the module as `Readline` (it's now `Gitsh::LineEditor`) and makes reference to supporting libedit (which we no longer support).

mike-burns: Good catch. Is it worth it to even leave this documentation here?

georgebrock:
> Do you know: does this behave well on a Ruby that was built against libedit? My understanding is that it will work fine: you're linking against Readline and ignoring whatever Ruby does.

That's definitely the intention, but I don't think I've explicitly tested it. I will do so, and report back.

ext/gitsh/src/line_editor.c
georgebrock: Deleting it is definitely one way to update it!

lib/gitsh/line_editor.rb
georgebrock: This was much easier than getting the C implementation of `delete_text` to work properly across Ruby version with multi-byte strings!

georgebrock:
Yep, this works fine with a Ruby version built against libedit.

georgebrock:
Given the high complexity of this, and how many related things there are, I decided to keep a larger number of distinct commits than I would for a typical PR, and to merge them with `--no-ff` to keep them grouped together in the Git history.

PR #257, opened 2016-07-02
Rename INSTALL to INSTALL.md

calebeby:


georgebrock:
Thanks for this. Unfortunately, this change will be more complex than just renaming the file.

The `INSTALL` file is generated by the `configure` script, based on `INSTALL.in`. To change it would involve renaming `INSTALL.in` to `INSTALL.md.in`, and updating the `AC_CONFIG_FILES` directive in `configure.ac`.

PR #259, opened 2016-07-23
Tab complete spaces (and other special characters)

georgebrock:
Allows users to tab-complete options that contain exotic characters like spaces, by:

* Explicitly telling GNU Readline which characters constitute word boundaries in the gitsh language.
* Using GNU Readline's `char_is_quoted_p` setting to inform Readline when a word boundary is escaped.
* Expanding the parser to support more escape sequences.
* Decorating the completer with a new `Gitsh::CompletionEscaper` which unescapes user input on the way in (so we can match it against the possibly completions) and then escapes the completions on the way out (so they can be correctly parsed by the shell)

### DIFF
--- a/lib/gitsh/completion_escaper.rb
+++ b/lib/gitsh/completion_escaper.rb
@@ -1,0 +1,46 @@
+require 'gitsh/parser'
+
+module Gitsh
+  class CompletionEscaper
+    ESCAPABLES = {
+      nil => Gitsh::Parser::UNQUOTED_STRING_ESCAPABLES,
+      '"' => Gitsh::Parser::SOFT_STRING_ESCAPABLES,
+      "'" => Gitsh::Parser::HARD_STRING_ESCAPABLES,
+    }.freeze
+
+    def initialize(completer, options)
+      @completer = completer
+      @line_editor = options[:line_editor]
+    end
+
+    def call(input)
+      completer.call(unescape(input)).map { |option| escape(option) }
+    end
+
+    private
+
+    attr_reader :completer, :line_editor
+
+    def escape(option)
+      escaped = option.gsub(/([#{escapables}])(?!$)/) { |char| "\\#{char}" }
+
+      if completing_quoted_argument?
+        escaped.strip
+      else
+        escaped
+      end
+    end
+
+    def unescape(input)
+      input.gsub(/\\([#{escapables}])/, '\1')
+    end
+
+    def escapables
+      ESCAPABLES[line_editor.completion_quote_character]
+    end
+
+    def completing_quoted_argument?
+      !line_editor.completion_quote_character.nil?
+    end
+  end
+end
--- a/lib/gitsh/completion_escaper.rb
+++ b/lib/gitsh/completion_escaper.rb
@@ -22,15 +22,15 @@ module Gitsh
     attr_reader :completer, :line_editor
 
     def escape(option)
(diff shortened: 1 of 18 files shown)


lib/gitsh/completion_escaper.rb
```diff
@@ -0,0 +1,83 @@
...
+          escape_chars(option, UNQUOTED_STRING_ESCAPABLES)
+        end
+      end
+
+      def escape_chars(text, chars)
+        quotable_characters = Regexp.escape(chars)
+        text.gsub(/([#{quotable_characters}])(?!$)/) { |char| "\\#{char}" }
+      end
+
+      def unescape(input)
```
georgebrock: This doesn't interpret escape sequences in the same way as the parser: we should only allow escape sequences for the characters listed in the relevant `*__STRING_ESCAPABLES` constant, any other `\` should by taken as a literal.

Basis of a test could be:

<pre><code>!touch file.txt
add f\i<kbd>tab</kbd>  # will complete to `add file.txt`, but shouldn't</code></pre>

georgebrock:
Regression caught by manual testing: something after 16ddce7e7b8c3f07345ce2cb0fcc45c2e183eea5 introduced a bug where the line `:set color.ui true` in my `.gitshrc` file causes an error. Seems to be because I changed a `[\s…]` to `[ …]` in the parser.

lib/gitsh/completion_escaper.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    private
+
+    attr_reader :completer, :line_editor
+
+    def escape(option)
+      escaped = option.gsub(/([#{escapables}])(?!$)/) { |char| "\\#{char}" }
+
+      if completing_quoted_argument?
```
georgebrock: This is a little unpleasant, and is refactored out in #260. I didn't include that refactoring in this PR because it touched a few different classes, and this PR is already pretty large.

lib/gitsh/interactive_runner.rb
```diff
@@ -41,7 +43,13 @@ def run
 
     def setup_line_editor
       line_editor.completion_append_character = nil
-      line_editor.completion_proc = Completer.new(line_editor, env)
+      line_editor.completion_proc = CompletionEscaper.new(
+        Completer.new(line_editor, env),
+        line_editor: line_editor,
+      )
+      line_editor.completer_quote_characters = '\'"'
```
georgebrock: How about using other syntax?

lib/gitsh/parser.rb
```diff
@@ -3,6 +3,11 @@
 
 module Gitsh
   class Parser < Parslet::Parser
+    UNQUOTED_STRING_TERMINATORS = %q(\\s'"&|;#).freeze
+    UNQUOTED_STRING_ESCAPABLES = (UNQUOTED_STRING_TERMINATORS + %q(\\\$)).freeze
+    SOFT_STRING_ESCAPABLES = %q(\\\$").freeze
+    HARD_STRING_ESCAPABLES = %q(\\\').freeze
```
georgebrock: All of this is terrible.

georgebrock:
@mike-burns Based on our conversation at the thoughtbot summit about how hard it was to understand the character class constants in `Gitsh::Parser`, what do you think of 36d999d539f55924c1c68a3340182b1af49d0513?

mike-burns:
1. I have, for many years, wished regexps were written like this, and therefore
2. that commit looks great!

PR #260, opened 2016-07-30
Refactor completion suffix

georgebrock:
### Before this commit

We'd set `Gitsh::LineEditor.completion_append_character` to `nil`, and then append a space to any completion option that wasn't a directory path (e.g., completions might include `"file.txt "` with a trailing space, and `"dir/"` without a trailing space).

The benefit of this was that you could tab complete a directory name, not have a space appended, and continue typing the path. You could also tab complete any other argument, have a space appended, and begin typing the next argument.

The down side was that we needed to manually remove the space from the end of the options if we were completing a quoted argument, and we needed to avoid escaping spaces at the end of a completion when completing an unquoted argument.

### After this commit

We set `LineEditor.completion_append_character` dynamically based on the completion options that matched the user input. If we have a single option that's a directory path, we set it to `nil`. If we have multiple options, or a single option that's not a directory path, we set it to space.

This maintains almost identical behaviour, but has a few advantages:

* We can drop all the special case code for not escaping the trailing space, or for removing the trailing space when completing a quoted argument.
* Readline will add a trailing space for us after the end of a completed quoted argument, which saves typing it by hand and is more similar to what other shells do.

(This PR depends on #259)

### DIFF
--- a/lib/gitsh/completer.rb
+++ b/lib/gitsh/completer.rb
@@ -9,7 +9,7 @@ module Gitsh
     end
 
     def call(input)
-      InputCompleter.new(input, @line_editor, @env, @internal_command).complete
+      InputCompleter.new(input, @line_editor, @env, @internal_command).call
     end
 
     class InputCompleter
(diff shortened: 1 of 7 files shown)


lib/gitsh/completer.rb
```diff
@@ -20,14 +20,27 @@ def initialize(input, line_editor, env, internal_command)
...
+        set_completion_append_character
+        matches
       end
 
       private
 
       attr_reader :input, :line_editor, :env, :internal_command
 
+      def set_completion_append_character
+        if matches.size == 1 && matches.first[-1] == '/'
```
georgebrock: `… && matches.first.end_with?('/')`

georgebrock: Extract into a setter method and a `completion_append_character` method; inline the setter method into `call`?

PR #264, opened 2016-07-31
Improve prompt generation speed

calleerlandsson:
Generating and displaying the prompt in gitsh is currently painfully slow on my MacBook Air. This PR is an attempt to improve on this.

I started to investigate by printing all git commands passed to `Open3.capture3` in `GitRepository#git_output`:

```
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git symbolic-ref HEAD --short"
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git symbolic-ref HEAD --short"
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git symbolic-ref HEAD --short"
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git status --porcelain"
"/usr/bin/env git status --porcelain"
"/usr/bin/env git config --get-color gitsh.color.modified yellow"
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git status --porcelain"
"/usr/bin/env git status --porcelain"
```

Generating the prompt on my system took ~1.1 seconds.

I observed that most of the calls to external git commands were duplicated.

I then added `puts` statements in many of the methods in `Prompter` and `PromptColor` in order to find out where the calls were made. I went on to reduce this number by introducing memoization in a way that, to my knowledge, doesn't affect the behavior of the program.

The three commits in this PR are the results of this. It brings the generation time down to 0.3 seconds and the following calls to external git commands:

```
"/usr/bin/env git rev-parse --git-dir"
"/usr/bin/env git status --porcelain"
"/usr/bin/env git config --get-color gitsh.color.modified yellow"
"/usr/bin/env git symbolic-ref HEAD --short"
```

In doing this I exposed a new `status` method on `GitRepository`. I didn't figure out a good way to test this new method. It is currently implicitly tested through the `has_untracked_files?` and `has_modified_files?` specs. I considered the following options:
1. Duplicate the existing tests but in the exercise, call `status.has_untracked_files?` instead of `has_untracked_files?`. I don't think the duplicated tests would provide much value.
2. Create a new spec for `GitRepository::Status` passing in example `git status --porcelain` output and replace the existing specs with specs that test that the original methods delegate to the status object. Since the existing specs tests the git repository on an integration level by altering a real git repository and the new specs would be hard to implement this way, this option seems to reduce the value of the test suite.

I'd love some input on this.

In my last commit I also inline the `PromptColor` method within the `Prompter` to make use of the cache inside it. Only having one method and no private helpers, it seemed like the `PromptColor` didn't really pull it's own weight so I think this is a good trade off. I'm also unsure of the value provided by the `PromptColor` spec since it mostly duplicated the conditional logic of the implementation. Therefor I didn't add any similar tests for `Prompter`.

### DIFF
--- a/lib/gitsh/git_repository.rb
+++ b/lib/gitsh/git_repository.rb
@@ -1,14 +1,12 @@
 require 'open3'
 require 'shellwords'
+require 'gitsh/git_repository/status'
 
 module Gitsh
   class GitRepository
-    def initialize(env)
+    def initialize(env, options={})
       @env = env
-    end
-
-    def initialized?
-      git_dir && File.exist?(git_dir)
+      @status_factory = options.fetch(:status_factory, Status)
     end
 
     def git_dir
@@ -19,12 +17,8 @@ module Gitsh
       current_branch_name || current_tag_name || abbreviated_sha
     end
 
-    def has_untracked_files?
-      status.untracked_files.any?
-    end
-
-    def has_modified_files?
-      status.modified_files.any?
+    def status
+      status_factory.new(git_output('status --porcelain'), git_dir)
     end
 
     def heads
@@ -97,7 +91,7 @@ module Gitsh
 
     private
 
-    attr_reader :env
+    attr_reader :env, :status_factory
 
     def current_branch_name
       branch_name = git_output('symbolic-ref HEAD --short')
@@ -120,34 +114,12 @@ module Gitsh
       end
     end
 
-    def status
-      StatusParser.new(git_output('status --porcelain'))
-    end
-
     def git_output(command)
       Open3.capture3(git_command(command)).first.chomp
     end
 
     def git_command(sub_command, force_default = false)
       "#{env.git_command(force_default)} #{sub_command}"
-    end
-
-    class StatusParser
-      def initialize(status_porcelain)
-        @status_porcelain = status_porcelain
-      end
-
-      def untracked_files
-        status_porcelain.lines.select { |l| l.start_with?('??') }
-      end
-
-      def modified_files
-        status_porcelain.lines.select { |l| l =~ /^ ?[A-Z]/ }
-      end
-
-      private
-
-      attr_reader :status_porcelain
     end
   end
 end
--- a/lib/gitsh/prompter.rb
+++ b/lib/gitsh/prompter.rb
@@ -16,83 +16,105 @@ module Gitsh
     end
 
     def prompt
-      padded_prompt_format.gsub(/%[bBcdDw#]/, {
-        '%b' => branch_name,
-        "%B" => shortened_branch_name,
-        '%c' => status_color,
-        '%d' => working_directory,
-        '%D' => File.basename(working_directory),
-        '%w' => clear_color,
-        '%#' => terminator
-      })
+      Prompt.new(env, use_color, prompt_color).to_s
     end
 
     private
 
-    attr_reader :env, :prompt_color
+    attr_reader :env, :use_color, :prompt_color
 
-    def working_directory
-      Dir.getwd.sub(/\A#{Dir.home}/, '~')
-    end
-
-    def padded_prompt_format
-      "#{prompt_format.chomp} "
-    end
-
-    def prompt_format
-      env.fetch('gitsh.prompt') { DEFAULT_FORMAT }
-    end
-
-    def shortened_branch_name
-      branch_name[0...BRANCH_CHAR_LIMIT] + ellipsis
-    end
-
-    def ellipsis
-      if branch_name.length > BRANCH_CHAR_LIMIT
-        '…'
-      else
-        ''
+    class Prompt
+      def initialize(env, use_color, prompt_color)
+        @env = env
+        @use_color = use_color
+        @prompt_color = prompt_color
       end
-    end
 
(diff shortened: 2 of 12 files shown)


lib/gitsh/git_repository.rb
```diff
@@ -8,7 +8,8 @@ def initialize(env)
     end
 
     def initialized?
-      git_dir && File.exist?(git_dir)
+      dir = git_dir
+      dir && File.exist?(dir)
```
georgebrock: What do you think of this as an alternative?

```ruby
def initialized?
  File.exist?(git_dir)
rescue TypeError
  false
end
```

lib/gitsh/prompter.rb
```diff
@@ -11,25 +10,33 @@ class Prompter
...
-      padded_prompt_format.gsub(/%[bBcdDw#]/, {
-        '%b' => branch_name,
-        "%B" => shortened_branch_name,
-        '%c' => status_color,
-        '%d' => Dir.getwd,
-        '%D' => File.basename(Dir.getwd),
-        '%w' => clear_color,
-        '%#' => terminator
-      })
+      reset_chached_repo_info
```
georgebrock: Typo: s/chached/cached/

georgebrock: We generally avoid creating state outside of `#initialize` methods in this codebase, and often create private classes that can hold the state they need to process a specific call (e.g. in the `Gitsh::Completer#call` method). Does it make sense to follow that pattern here too?

Something like this maybe?

```ruby
def prompt
    Prompt.new(
      env: env,
      repo_initialized: …,
      branch_name: …,
      repo_status: …,
    ).to_s
end
```

georgebrock:
Thanks so much for this! The speed improvement is great.

> In doing this I exposed a new `status` method on `GitRepository`. I didn't figure out a good way to test this new method. It is currently implicitly tested through the `has_untracked_files?` and `has_modified_files?` specs. I considered the following options:
>
> 1. Duplicate the existing tests but in the exercise, call `status.has_untracked_files?` instead of `has_untracked_files?`. I don't think the duplicated tests would provide much value.
> 2. Create a new spec for `GitRepository::Status` passing in example `git status --porcelain` output and replace the existing specs with specs that test that the original methods delegate to the status object. Since the existing specs tests the git repository on an integration level by altering a real git repository and the new specs would be hard to implement this way, this option seems to reduce the value of the test suite.
>
> I'd love some input on this.

I prefer the second option. As long as there's a true integration spec somewhere that exercises a couple of the prompt variants, I think it's OK to test this at the unit level with stubbed collaborators. [`spec/integration/prompt_spec.rb`](https://github.com/thoughtbot/gitsh/blob/master/spec/integration/prompt_spec.rb) seems to cover the basics already. Maybe we just need another example in there to demonstrate the sigil changing when we modify files?

> In my last commit I also inline the `PromptColor` method within the `Prompter` to make use of the cache inside it. Only having one method and no private helpers, it seemed like the `PromptColor` didn't really pull it's own weight so I think this is a good trade off. I'm also unsure of the value provided by the `PromptColor` spec since it mostly duplicated the conditional logic of the implementation. Therefor I didn't add any similar tests for `Prompter`.

I think there was some value in these specs. Particularly, the order of the `elsif` cases matters, and without that test I don't think we exercise all of the possibilities anywhere anymore.

How would you feel about changing the `Gitsh::PromptColor` class to accept a `Status` object as an argument to its `#status_color` method? That way, it could still be tested separately, but it could also use the cached state.

lib/gitsh/git_repository.rb
mike-burns: I'm confused in all these suggestions -- not sure what I'm missing -- but what is the `TypeError` that would be raised? Can `git_output` produce `nil`?

georgebrock: The `git_dir && …` in the original implementation appeared to be there to guard against `nil`, so I assumed that it could.

However, it seems I'm wrong. It can be the empty string, but since `File.exist?('')` returns `false` we can probably simplify this even further.

calleerlandsson:
I'm sorry for dropping the ball on this one.

I have now rebased on master and updated the PR according to your comments:

- I added specs for `Gitsh::GitRepository::Status` and a sigil example to `spec/integration/prompt_spec.rb`
- I brought back `PromptColor` and its specs and made `status_color` accept a `Status` object as its parameter.
- I extracted a `Prompt` class encapsulating the cached state required for prompt generation.
- Simplified the implementation of `#initialized?`.

I also:

- Moved `GitRepository#initialized?` into `GitRepository::Status` to limit the number of parameters to pass to `PromptColor#status_color`.
- Realized that the `GitRepository` methods now delegating to the `Status` object and the `Environment` methods delegating to these were used only by objects with access to a `Status` object. Because of this, I removed the delegating methods.

georgebrock:
This is great, thanks Calle!

PR #265, opened 2016-08-06
Ignore shared objects

srstevenson:
This causes Git to ignore `ext/gitsh/line_editor_native.so`, which was created when building `gitsh` on Linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ lib/gitsh/version.rb
 gitsh-*.tar.gz
 vendor/gems
 *.o
+*.so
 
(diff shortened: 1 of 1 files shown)


georgebrock:
Good catch, thanks. I mostly develop on OS X which creates `.bundle` files instead of `.so` files.

georgebrock:
Merged as d2ed25825e134057ad044f2edf8e1bf3feee48be — I shuffled the order slightly so the `.so` rule was next to the `.bundle` rule, but you retain authorship on the commit.

PR #266, opened 2016-08-06
Show home directory as a tilde in the prompt

srstevenson:
When `gitsh` is launched from the user's home directory, the basename of the current working directory shown in the prompt is typically the same as the user's username, which is a potential source of confusion as to what this prompt section shows. With this commit, the directory name is shown as a tilde when the current working directory is the user's home directory, as is conventional for Unix shells.

Without this commit:

    $ cd ~
    $ pwd
    /home/srs
    $ gitsh
    gitsh 0.10
    Type :exit to exit
    srs uninitialized!!

With this commit:

    $ cd ~
    $ pwd
    /home/srs
    $ gitsh
    gitsh 0.10
    Type :exit to exit
    ~ uninitialized!!

### DIFF
--- a/lib/gitsh/prompter.rb
+++ b/lib/gitsh/prompter.rb
@@ -20,8 +20,8 @@ module Gitsh
         '%b' => branch_name,
         "%B" => shortened_branch_name,
         '%c' => status_color,
-        '%d' => Dir.getwd,
-        '%D' => File.basename(Dir.getwd),
+        '%d' => working_directory,
+        '%D' => File.basename(working_directory),
         '%w' => clear_color,
         '%#' => terminator
       })
@@ -30,6 +30,10 @@ module Gitsh
     private
 
     attr_reader :env, :prompt_color
+
+    def working_directory
+      Dir.getwd.sub(/\A#{Dir.home}/, '~')
+    end
 
     def padded_prompt_format
       "#{prompt_format.chomp} "
--- a/spec/units/prompter_spec.rb
+++ b/spec/units/prompter_spec.rb
@@ -118,14 +118,16 @@ describe Gitsh::Prompter do
         env = env_double(format: '%d')
         prompter = Gitsh::Prompter.new(env: env)
 
-        expect(prompter.prompt).to eq "#{Dir.getwd} "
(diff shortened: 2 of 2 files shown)


lib/gitsh/prompter.rb
```diff
@@ -31,6 +31,10 @@ def prompt
 
     attr_reader :env, :prompt_color
 
+    def cwd_basename
+      Dir.getwd == Dir.home ? '~' : File.basename(Dir.getwd)
```
mike-burns: How about:

```ruby
File.basename(Dir.getwd).sub(/\A#{Dir.home}/, '~')
```

so that we always get tilde for homedir?

srstevenson: Do you mean to do the replacement in the `%d` format string too? i.e. something like:
```diff
diff --git a/lib/gitsh/prompter.rb b/lib/gitsh/prompter.rb
index b650cb3..a977700 100644
--- a/lib/gitsh/prompter.rb
+++ b/lib/gitsh/prompter.rb
@@ -20,8 +20,8 @@ module Gitsh
         '%b' => branch_name,
         "%B" => shortened_branch_name,
         '%c' => status_color,
-        '%d' => Dir.getwd,
-        '%D' => File.basename(Dir.getwd),
+        '%d' => working_directory,
+        '%D' => File.basename(working_directory),
         '%w' => clear_color,
         '%#' => terminator
       })
@@ -31,6 +31,10 @@ module Gitsh
 
     attr_reader :env, :prompt_color
 
+    def working_directory
+      Dir.getwd.sub(/\A#{Dir.home}/, '~')
+    end
+
     def padded_prompt_format
       "#{prompt_format.chomp} "
     end
```

mike-burns: Aha, good point. Yes, I'd like that.

srstevenson:
I've changed this to replace the home directory with a tilde in both the `%D` and `%d` format strings based on @mike-burns's feedback, and updated the tests to reflect this.

georgebrock:
Thanks! Merged as d1e59e2d65ae660c992b35aa0a1f36811d6b1322 to keep this history linear.

PR #267, opened 2016-08-26
Don't always close quotes when tab completing.

### LINKED ISSUE #262: Tab completing quoted paths is awkward
Tab completing a quoted path is awkward, e.g. in this repository <kbd>add "li<kbd>tab</kbd></kbd> would complete to <samp>add "lib/"</samp>, and to continue either typing or tab completing the path would require deleting the trailing `"`.

georgebrock:
When the thing we've managed to complete isn't a full argument (i.e. it's a directory name, and the user may want to add more components to the path), don't append a quote if it's a quoted argument.

Fixes #262


mike-burns:
LGTM.

PR #269, opened 2016-09-02
A non-recursive Makefile

mike-burns:
Recursive Makefiles are [less ideal](http://aegis.sourceforge.net/auug97.pdf) than iterative ones, for quite a few reasons. But the main driver for this move is as a simplification so that we can concentrate on the `ext` directory separately from autotools.

This PR does not completely fix the `distcheck` target, but `dist` and `check` each work.


Makefile.am
```diff
@@ -1,6 +1,58 @@
...
+	lib/gitsh/version.rb.in
+
+bin_PROGRAMS = gitsh
+gitsh_SOURCES = src/gitsh.c
+AM_CPPFLAGS = -DGITSH_RB_PATH="\"$(rubydir)/$(PACKAGE).rb\""
+
+ruby_SCRIPTS = src/gitsh.rb
+CLEANFILES = $(ruby_SCRIPTS) lib/gitsh/version.rb
+
+edit = sed \
```
mike-burns: It probably makes sense to re-order this file.

georgebrock:
@mike-burns Thanks for this! It was much closer to producing a working distribution than master.

I did find a couple of problems with the tarballs built by `make dist` (which `make distcheck` would've caught, if we could use it), but I think I've solved them:

* Several install locations changed, because the path to the source files relative to the Makefile changed, and we use the `nobase_*` prefix in a few places. Rather than fighting Automake, I updated things to load the files from their new locations (c47bd5c).
* We weren't installing the built line editor extension anywhere.

If you have time for a quick look at the changes, I'd appreciate it.

mike-burns:
I like it! Those file moves shouldn't be an issue for package maintainers -- it will work fine for the OpenBSD package, at least.

Didn't try it locally, but if it works them it LGTM.

georgebrock:
I built a tarball from this branch, which I could install successfully on both OS X and Ubuntu.

Issue #177, opened 2014-07-10
Strange history display in the prompt


jlecour:
On Mac OS X (10.9.4), with Gitsh 0.7 I have a strange display of history.

I would have made a nice animated gif if I knew how, but I don't so here is a _transcript_ : 

```
~/Sandbox/rails/rails → gitsh
gitsh 0.7
Type :exit to exit
```

Then I type a few Git commands to populate my history

```
rails 5f844d6...@ status
HEAD detached at 5f844d6
nothing to commit, working directory clean
rails 5f844d6...@ fetch
```

Then if I type :arrow_up: i get `rails 5f844d6...@ fetch` which is OK.

But if I type again :arrow_up: I get `rails 5f844d6...@ fetch            status`

The line is off by 17 characters, starting by the previous command and padded with spaces.

Then if I type :arrow_left: multiple times, I can't go back to the beginning of the prompte, it stops at the 17 characters block.

Here is the final screenshot since the copy/past doesn't show the proper padding space (probably not regular space chars)

![gitsh_prompt](https://cloud.githubusercontent.com/assets/5142/3536536/f933ab2c-0812-11e4-8987-67a045ee43a2.png)

PR #273, opened 2016-12-09
Add install ruby step to mac os installation guide

wdspkr:
If a mac user followed the standard installation guide and had the standard mac os ruby installed he would have a very ugly and almost unworkable history behaviour (see https://github.com/thoughtbot/gitsh/issues/177).

This adds the simple solution for this issue to the install guide, so other users don't need to dive into the github issue history to get a nice gitsh experience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ sh$</code></pre>
         gitsh% ⏎
         # On branch master
         nothing to commit, working directory clean
-        gitsh% 
+        gitsh%
 
 * Easily execute shell commands:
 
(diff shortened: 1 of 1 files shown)


georgebrock:
Thanks for this.

Since #256 gitsh's used a custom Readline integration, which means we can avoid these problems regardless of the line editor that Ruby was compiled with.

I'm working on a release today, and once that's out the extra `brew install ruby` step won't be required.

wdspkr:
Sweet! Thanks for the great work.

PR #275, opened 2016-12-22
Use SHA256 for Homebrew and Arch package digests.

georgebrock:
The package managers use the digest to verify the integrity of the tarball they've downloaded. 

Homebrew has deprecated SHA1, and we were using the same digest in both places, so upgrade both for simplicity.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -122,4 +122,4 @@ release_clean_arch:
 	rm -rf release-arch
 
 DIST_SHA: Makefile dist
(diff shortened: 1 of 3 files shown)


georgebrock:
I'm going to try using this for the 0.11 release, and if the Homebrew formula works, then I'm going to merge it.

georgebrock:
Yep, the updated formula worked fine for v0.11. Merging.

Issue #261, opened 2016-07-30
Tab completion only works for the first command on the line


georgebrock:
It's possible to tab complete the first command on a line, e.g. <kbd>ad<kbd>tab</kbd></kbd> will complete to <samp>add </samp>.

It's not possible to complete subsequent commands, e.g. <kbd>add thing.txt && comm<kbd>tab</kbd></kbd> won't complete to <samp>add thing.txt && commit </samp>, but it might complete to something else if `com` matches the name of a file, remote, or branch.

This is because command completion or argument completion is selected based on the naive assumption that if any spaces appear before the thing we're completing, it must be an argument.

georgebrock:
One solution would be to pass the user's input through `Gitsh::Parser` to determine exactly what we're completing. We'd have to close unclosed quotes to make sure the input was valid.

This would also give us a good starting point for some more advanced context aware completion (e.g., #25, #80, #220)

georgebrock:
Closing in favour of #296.

Issue #220, opened 2015-01-20
Auto complete stash commands


lode:
I miss auto completion for `stash` commands.

Mainly the stash names `stash@{0}` when typing `stash show -p` + tab I would like it to expand to `stash show -p stash@{` and to show me a list of stashes.

But also the main commands would be nice, like `stash a` + tab => `stash apply`.

Right now branch names and directories are auto completed and shown on multiple choices. I don't know about other people, but I hardly use those functionality with `stash`.

georgebrock:
After the tab completion improvements introduced in #296, #304, #310, #311, #312, #313, we can now tab complete stash commands (and other popular commands with sub-commands!).

Issue #80, opened 2014-02-11
Add tab completion for options


JoelQ:
According to the README:

> Tab completion for git commands, aliases, and branches without modifying your shell settings, and without any extra setup for aliases and third party git commands.

Tab completion works for commands/branches but not for options. For example, if I want to run `rebase --continue` I can tab-complete the `rebase` but not the `--continue`. ZSH allowed me to complete both. It would be nice to add this ability to `gitsh`.

georgebrock:
After the tab completion improvements introduced in #296, #304, #310, #311, #312, #313, we can now tab complete options!

Issue #35, opened 2013-12-07
Add a :function command for defining functions


georgebrock:
For example:

```
:function pair name nick {
  :set user.name "George Brocklehurst and $name"
  :set user.email "support+george+$nick@thoughtbot.com"
  :set gitsh.prompt "$nick+$gitsh.prompt"
}

# usage:
# gitsh@ @pair "Mike Burns" mburns
# mburns+gitsh@
```

PR #279, opened 2016-12-30
Replace Parslet with RLTK

georgebrock:
This is a really major change, and therefore a really big PR. Strap in.

## What changed?

This PR replaces the [Parslet][1]-based parsing expression grammar (PEG) with an [RLTK][2]-based context free grammar (CFG).

The `Gitsh::Parser` (which converted text into a AST) and `Gitsh::Transformer` (which converted the AST into useful domain objects) have been replaced with a `Gitsh::Lexer` (which converts text into tokens) and `Gitsh::Parser` (which converts tokens into useful domain objects).

[1]: https://github.com/kschiess/parslet
[2]: https://github.com/chriswailes/RLTK

## Why?!

Distinguishing between _invalid input_ and _valid but incomplete input_ is really hard with a PEG. There are various papers on improving PEG error reporting, and some support in Parslet for various strategies, but I couldn't find a satisfactory way to accurately work out why parsing failed without making the grammar significantly more complex.

We want to add support for various multi-line constructs, which means we need to know if a line is invalid or just waiting for more input. The specific issues I have in mind are:

* Support multi-line strings (#31)
* Add a `:function` command for defining functions (#35)

We also want to improve tab-completion in various ways, which also involves working with very similar types of incomplete input. Most of the enhancements we have in mind will only be possible if we know more exactly what we're completing, and re-using some of the parsing code to understand the input seemed like the most logical approach. Specific issues are:

* Support the same tab completion options as zsh (#25)
* Add tab completion for options (#80)
* Auto complete stash commands (#220)
* Tab completion only works for the first command on the line (#261)

## Still to do

This PR isn't 100% complete, but I wanted to open it to get early feedback.

* [x] Fix adjacent subshells are parsed incorrectly (e.g. `:echo $(:echo foo)$(:echo bar)`)
* [x] Investigate caching the parser definition either when building the distribution, or during installation (https://github.com/chriswailes/RLTK#finalization-options)


lib/gitsh/lexer.rb
```diff
@@ -0,0 +1,101 @@
...
+    rule(/"/, :soft_string) { pop_state }
+
+    [:default, :soft_string].each do |state|
+      rule(/\$[a-z_][a-z0-9_.-]*/i, state) { |t| [:VAR, t[1..-1]] }
+      rule(/\$\{[a-z_][a-z0-9_.-]*\}/i, state) { |t| [:VAR, t[2..-2]] }
+    end
+
+    [:default, :soft_string].each do |state|
+      rule(/\$\(/, state) do
+        @subshell_parens = 1
```
mike-burns: Should this be `+=` instead of `=`? e.g. `$( foo "$(bar)")`

lib/gitsh/lexer.rb
```diff
@@ -0,0 +1,101 @@
...
+    def self.lex(string, file_name = nil, env = RLTK::Lexer::Environment.new(@start_state))
+      tokens = super
+
+      if env.state == :hard_string
+        tokens.insert(-2, RLTK::Token.new(:MISSING, '\''))
+      elsif env.state == :soft_string
+        tokens.insert(-2, RLTK::Token.new(:MISSING, '"'))
+      elsif env.state == :subshell
+        tokens.insert(-2, RLTK::Token.new(:MISSING, ')'))
+      end
```
mike-burns: Could use a `case` statement.

lib/gitsh/lexer/character_class.rb
```diff
@@ -13,13 +13,13 @@ def +(other)
...
 
-      def parser_atom
-        Parslet.match(to_regexp)
-      end
-
       def to_regexp
         Regexp.new("[#{Regexp.escape(characters.join)}]")
       end
+
+      def to_negative_regexp
```
mike-burns: I like it.

spec/units/commands/git_command_spec.rb
```diff
@@ -53,5 +59,55 @@
         env,
       )
     end
+
+    context 'with autocorrect enabled' do
```
mike-burns: Is this new functionality? Or, I guess, why is this test here? Was this handled in the parser before?

lib/gitsh/lexer.rb
georgebrock: This rule puts us in the `subshell` state, so this is setting up the counter for a new subshell. Once we're in that state we do increment this when we encounter any `(` character.

lib/gitsh/lexer.rb
georgebrock: Good call.

spec/units/commands/git_command_spec.rb
georgebrock: Yes, this was handled in the old parser. I couldn't work out how to handle this in the new parser so this seemed like the next best thing.

mike-burns:
What will this do to `vendor/`? Will RLTK fit nicely in there?

It looks like it will but it's hard to know for sure, so just raising it as a thing to consider.

georgebrock:
> What will this do to vendor/? Will RLTK fit nicely in there?

It will, yes. RLTK is much bigger than Parslet, but not so big that I'm worried about it.

georgebrock:
> Investigate caching the parser definition either when building the distribution, or during installation (https://github.com/chriswailes/RLTK#finalization-options)

I looked into this, and decided against it for two reasons:

1. Loading the grammar doesn't appear to be particularly slow. It's quite a simple grammar.
2. The caching uses `Marshal.dump`, so it might not work well across Ruby versions. That'd mean building the cached version at install time rather than packaging time, which seems complicated.

spec/units/token_parser_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+      result = described_class.parse(tokens([:WORD, 'commit'], [:EOS]))
+
+      expect(result).to eq command
+      expect(Gitsh::Commands::Factory).to have_received(:build).with(
+        Gitsh::Commands::GitCommand,
+        hash_including(command: 'commit', args: []),
+      )
+    end
+
+    it 'parses Git commands broken into multiple words' do
```
sharplet: Is this to support dynamic commands built from variables?

spec/units/token_parser_spec.rb
```diff
@@ -173,6 +173,17 @@
         env: env, command: 'commit', args: [],
       )
     end
+
+    it 'parses blank lines' do
+      expect(parse(tokens([:EOS]))).to be_a Gitsh::Commands::Noop
+      expect(parse(tokens([:SPACE], [:EOS]))).to be_a Gitsh::Commands::Noop
+    end
+
+    it 'does not parse just a semicolon' do
```
sharplet: What do you think about allowing this? I just tested it and zsh does:

```
% ;
% ; echo post
post
% echo pre;
pre
```

sharplet: Although I suppose it's complicated by the fact that a noop usually means executing a default command, so maybe this is the right approach.

georgebrock: Both bash and sh consider `;` to be a syntax error. Combined with the default command considerations you mentioned, I think that's reason enough to keep the current behaviour.

spec/units/lexer_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+
+    it 'recognises single-quoted strings as single words' do
+      expect('\'foo bar\'').
+        to produce_tokens ['WORD(foo bar)', 'EOS']
+    end
+
+    ['\\', '\''].each do |char|
+      it "recognises single-quoted strings containing an escaped #{char.inspect}" do
+        expect("'foo\\#{char}bar'").
+          to produce_tokens ['WORD(foo)', "WORD(#{char})", 'WORD(bar)', 'EOS']
```
georgebrock: The job of the lexer is to convert syntax (`\$`) into tokens that are useful in the language (`WORD($)`). The way lexers typically do this is by testing the beginning of the input against a bunch of regular expressions, and going with whichever produces the longest match.

It's be very hard (maybe even impossible?) to write a single regular expression that could handle all of the possible variations of escaping (there's `\$x` to `WORD($x)`, but also more complex things like `\\$x` to `WORD(\) VAR(x)`). Even if it were possible, I think the current set of rules is clearer than whatever we'd need to produce a single `WORD` token for these cases.

There are also other reasons for the parser to combine multiple adjacent `WORD` tokens (see my answer to your comment below about parsing commands broken into multiple words), so it doesn't matter much if the lexer produces `WORD(x) WORD(y)` or `WORD(xy)`.

spec/units/token_parser_spec.rb
georgebrock: The intention behind this was to support that most general-purpose shells support, where you can mash a quoted word and an unquoted word together, and they're interpreted as a single word. For example, in bash, sh, and zsh, you can run `"l"s` or `ls` or `l's'` and they're all parsed the same way.

georgebrock:
Talked to @sharplet IRL about his subshell questions, we identified a couple of problems:

- It's possible to break the parser with things like `:echo $(:echo ")))")`.
- Parse errors in subshells are handled at the wrong time, e.g. `:echo foo $(;) bar` outputs a parse error, but still runs the command and outputs `foo  bar`.

We came up with a plan to solve both problems:

1. Add lexer support for parens as first-class operators (#278).
2. Get rid of the lexer's `:subshell` state, and just lex subshell contents as normal in the `:default` state.
3. Maintain a stack of closing tokens: when we encounter `(` we emit `LEFT_PAREN` and push `RIGHT_PAREN`, when we encounter `$(` we emit `SUBSHELL_START` and push `SUBSHELL_END`, and when we encounter `)` we emit whatever we pop from the stack (or default to `RIGHT_PAREN` if the stack is empty).
4. Modify the parser rules for subshells to something like `SUBSHELL_START program SUBSHELL_END`.
5. Modify the `Gitsh::Arguments::Subshell` class to compose a `Gitsh::Command::*`object instead of a `String`. This will involve changing the various command classes so that they can be passed an environment at execution time instead of initialisation time.

georgebrock:
Updated with 3 new commits, that implement the plan in my last comment to fix all the subshell problems @sharplet found.

/cc @sharplet @mike-burns

spec/units/lexer_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+    it 'does not interpret all \ characters in single-quoted arguments as escapes' do
+      ['a', ' ', '"', '&', '|', ';', '#', '$'].each do |char|
+        expect("'\\#{char}'").
+          to produce_tokens ['WORD(\\)', "WORD(#{char})", 'EOS']
+      end
+    end
+
+    it 'recognises single-quoted strings containing variable-like values' do
+      expect("'hello $world'").
```
sharplet: I wonder if it's worth a test for nested types of string? Say

```ruby
    it 'does not interpret single quotes in soft strings as hard strings' do
      expect(%q{"'$foo'"}).
        to produce_tokens ["WORD(')", "VAR(foo)", "WORD(')", "EOS"]
    end
```

spec/integration/subshell_spec.rb
```diff
@@ -31,4 +31,22 @@
...
+  it 'supports adjacent subshells' do
+    GitshRunner.interactive do |gitsh|
+      gitsh.type ':echo $(:echo foo)$(:echo bar)'
+
+      expect(gitsh).to output_no_errors
+      expect(gitsh).to output /\bfoo\nbar\b/
+    end
+  end
+
+  it 'is not confused by quoted parens in a subshell' do
```
sharplet: Lovely.

lib/gitsh/lexer.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+    rule(/\s*\(\s*/) do
+      push_state(:default)
+      right_paren_stack.push(:RIGHT_PAREN)
+      :LEFT_PAREN
+    end
+    rule(/\s*\)\s*/) do
+      pop_state
+      right_paren_stack.pop || :RIGHT_PAREN
```
georgebrock: I think it's preferable to leave this up to the parser. I see the lexer's responsibility as emitting tokens based on the string, and the parser's responsibility as deciding what those tokens mean (if they mean anything at all).

The `MISSING` token is a bit of a hack, which I'd remove if I could think of a better way of communicating that the lexer was finishing its run in a non-default state.

spec/units/lexer_spec.rb
georgebrock: Added. Thanks.

georgebrock:
Pretty sure this is ready to merge, but I want to resolve #284, #285, and #287 first so I can make a minor release.

PR #280, opened 2017-01-06
Refactor from runners to input strategies

georgebrock:
## What changed?

This PR replaces the `Gitsh::InteractiveRunner` and `Gitsh::ScriptRunner` classes with input strategies (`Gitsh::InputStrategies::Interactive` and `Gitsh::InputStrategies::File`). Previously each runner class's `#run` method would loop over available input and pass each command to the interpreter. Now the interpreter controls the loop and asks the strategy for commands.

The old `Gitsh::ScriptRunner` class was used to load the `.gitshrc` file at the start of an interactive session, and by the internal `:source` command. This has been replaced with a `Gitsh::FileRunner` that handles setting up a `Gitsh::InputStrategies::File` and a `Gitsh::Interpreter` for a given file.

Similarly, the old `Gitsh::Interpreter#execute` method was used by `Gitsh::Arguments::Subshell` to run the contents of a subshell. This has been replaced with a `Gitsh::StringRunner` and associated `Gitsh::InputStrategies::String`.

## Why?!

The main motivation was to support multi-line constructs. We would've needed to add very similar code for handling incomplete input to both the runner classes (once #279 is merged and we _can_ detect incomplete input). Now that the interpreter is responsible for controlling to overall flow of the session, we'll be able to handle incomplete input in one place.

If that weren't reason enough, there are a bunch of other nice things about this:

* It simplifies `Gitsh::CLI`, by conditionally creating strategy objects, instead of conditionally calling different runners.
* It makes the similarities that already existed between the two runner classes much more clear than they were before.
* It makes the process of running different types of input much easier (e.g. if we wanted to support a sh-like `gitsh -c status`, all we'd need to do now is get the `Gitsh::CLI` to use a different input strategy).


lib/gitsh/input_strategies/file.rb
```diff
@@ -0,0 +1,46 @@
...
+        @file = open_file
+      rescue Errno::ENOENT
+        raise NoInputError, "#{path}: No such file or directory"
+      rescue Errno::EACCES
+        raise NoInputError, "#{path}: Permission denied"
+      end
+
+      def teardown
+        if file
+          file.close
```
mike-burns: Do we have `#try`?

lib/gitsh/interpreter.rb
```diff
@@ -3,21 +3,31 @@
...
-    def initialize(env, options={})
-      @env = env
+    def initialize(options)
+      @env = options.fetch(:env)
       @parser_factory = options.fetch(:parser_factory, Parser)
+      @input_strategy = options.fetch(:input_strategy)
     end
 
+    def run
+      input_strategy.setup
```
mike-burns: Should we check a return value here?

Update: I see, we're doing that in `Interpreter#run`.

lib/gitsh/arguments/subshell.rb
```diff
@@ -1,22 +1,22 @@
...
     class Subshell
       def initialize(command, options = {})
         @command = command
-        @interpreter_factory = options.fetch(:interpreter_factory)
       end
 
       def value(env)
         capturing_env = CapturingEnvironment.new(env.clone)
-        interpreter_factory.new(capturing_env).execute(command)
+        StringRunner.run(env: capturing_env, command: command)
```
georgebrock: We're doing this pretty inconsistently (e.g. the previous line calls `CapturingEnvironment.new` directly). I'm not sure if it's been worthwhile, especially because we mostly inject dependencies in tests but rely on defaults in production code.

lib/gitsh/input_strategies/file.rb
georgebrock: We do not, it's an `ActiveSupport::CoreExt` thing.

lib/gitsh/interpreter.rb
georgebrock: I'm not totally sure what you mean here. This is `Interpreter#run`. Which return value should we be checking?

mike-burns: Ah whoops, I meant `CLI`, not `Interpreter`. Regardless: this line of code looks great.

PR #284, opened 2017-04-03
Support alternative names for Readline.

georgebrock:
The OpenBSD port for Readline calls the library `ereadline`, which means that the gitsh configure script needed to be patched to be used on OpenBSD systems.

This commit introduces a `READLINE_LIB` environment variable, which can be used to specify an alternative name to be passed to the Ruby C-extension's configuration script.

Based on @mike-burns' patch to `configure` in the OpenBSD ports tree:
  https://marc.info/?l=openbsd-ports&m=148340820830595&w=2

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -62,6 +62,14 @@ library is named ereadline, you might use:
           LDFLAGS="-L/usr/local/lib" \
           ./configure
 
+Finally, on macOS systems with a universal Ruby binary (e.g. /usr/bin/ruby),
+and a non-universal Readline shared library, you may need to explicitly pass
+an architecture to use via the $READLINE_ARCH environment variable:
+
+        READLINE_ARCH=x86_64 \
+          RUBY=/usr/bin/ruby \
+          ./configure
+
 
 Installation
 ------------
--- a/INSTALL
+++ b/INSTALL
@@ -53,6 +53,15 @@ you might use:
           LDFLAGS="-L/usr/local/opt/readline/lib" \
           ./configure
 
+If your system uses a different name for Readline, you can set the
+$READLINE_LIB environment variable. For example, on OpenBSD where the Readline
+library is named ereadline, you might use:
+
+        READLINE_LIB="ereadline" \
+          CPPFLAGS="-I/usr/local/include/ereadline" \
+          LDFLAGS="-L/usr/local/lib" \
+          ./configure
+
 
 Installation
 ------------
--- a/INSTALL.in
+++ b/INSTALL.in
@@ -62,6 +62,14 @@ library is named ereadline, you might use:
           LDFLAGS="-L/usr/local/lib" \
           ./configure
 
+Finally, on macOS systems with a universal Ruby binary (e.g. /usr/bin/ruby),
+and a non-universal Readline shared library, you may need to explicitly pass
+an architecture to use via the $READLINE_ARCH environment variable:
+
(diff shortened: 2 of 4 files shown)


georgebrock:
@mike-burns Three questions about this:

- Does the example I've given in `INSTALL.in` work on OpenBSD?
- Can you set install time environment variables in your port? Seems like adding them to `CONFIGURE_ENV` might do the trick.
- Are all of the environment variables in that example needed? The `LDFLAGS` setting is possibly redundant.

If so, I think that will remove the need for the `extconf.rb` related parts of your patch to `configure`. If no, we can expand this to include some OS-specific code in the configure script via [`AC_CANONICAL_TARGET`][1].

[1]: https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Canonicalizing.html

mike-burns:
Without `--with-readline-include=/usr/local/include/ereadline` it cannot find the right header, leading to this:

```
checking for rl_completion_suppress_quote in stdio.h,readline/readline.h,readline/history.h... no
```

And then it stops there.

Without `--with-readlinelib=ereadline` (which is a separate flag from `--with-readline-lib`), it cannot find the library at runtime:

```
 gitsh
ruby23:/usr/local/share/gitsh/ruby/lib/gitsh/line_editor_native.so: undefined symbol 'rl_completion_suppress_quote'
ruby23:/usr/local/share/gitsh/ruby/lib/gitsh/line_editor_native.so: undefined symbol 'rl_completion_quote_character'
/usr/local/share/gitsh/ruby/lib/gitsh/line_editor.rb:1:in `require': Cannot load specified object - /usr/local/share/gitsh/ruby/lib/gitsh/line_editor_native.so (LoadError)
        from /usr/local/share/gitsh/ruby/lib/gitsh/line_editor.rb:1:in `<top (required)>'
        from /usr/local/share/gitsh/ruby/lib/gitsh/environment.rb:3:in `require'
        from /usr/local/share/gitsh/ruby/lib/gitsh/environment.rb:3:in `<top (required)>'
        from /usr/local/share/gitsh/ruby/lib/gitsh/cli.rb:2:in `require'
        from /usr/local/share/gitsh/ruby/lib/gitsh/cli.rb:2:in `<top (required)>'
        from /usr/local/share/gitsh/ruby/gitsh.rb:6:in `require'
        from /usr/local/share/gitsh/ruby/gitsh.rb:6:in `<main>'
```

---

Surprisingly to me, the `libfiles` modification in my patch is no longer needed. Unsure what's happening there.

mike-burns:
Oh I didn't answer this one:

> Can you set install time environment variables in your port? Seems like adding them to CONFIGURE_ENV might do the trick.

Yup!

georgebrock:
I must be missing something here: setting `READLINE_LIB` should cause `--with-readlinelib` to be passed, and as far as I can tell from a quick reading of `mkmf.rb`, passing `--with-readline-include=x` just appends `-Ix` to `CPP_FLAGS`.

Maybe we can look at this together sometime?

georgebrock:
On a fresh install of OpenBSD 6.1, with this branch checked out, the example in `INSTALL` worked for me without any further changes to `configure` (I also had to set `RUBY=$(which ruby24)`, but that's easy enough to fix).

@mike-burns Were you using this branch when you tried it?

mike-burns:
I had manually applied the diff, so I likely missed something. If it works for you then A+, merge it!

PR #285, opened 2017-04-03
Improve configure script's source file discovery.

georgebrock:
- Uses Autoconf's `$srcdir` instead of `basename $0` as the root of the source tree.
- Removes a redundant `cut`. This was intended to strip `./` from the beginnings of the resulting paths, but it wasn't portable (`./` wasn't prepended to the paths on some systems), and the way this list is generated and used has changed so much that it no longer appears to matter if the `./` is there at all.

Partly based on Mike Burns' patch to `configure` in the OpenBSD ports tree:
  https://marc.info/?l=openbsd-ports&m=148340820830595&w=2


mike-burns:
That's a great improvement!

Issue #242, opened 2015-10-16
Thoughts on having a bare `cd` go to repo root


christoomey:
Using `gitsh` this past week I ended up using the internal `:cd` command to move into a subfolder in my repo to do some work there, then wanted to return to the repo root (in the `gitsh` context).

I was only able to do this by knowing the correct count of `../`s to get back to the root.

I'd like to add support for `:cd` with no argument to chdir to the repo root (much like a bare `cd` in the shell going to `$HOME`). I began work on this but hit a few snags and wanted to see what the team thought.
### Specific questions I have are:
- Does this make sense? Specifically, is it logical to think of a `gitsh` instance existing within the context of a single repo? This is exclusively how I use it, but might not be the case.
- If this does makes sense, what is the best implementation? I see two main options:
  - Dynamically find the root when `:cd` is invoked (using `git rev-parse --show-cdup`)
  - Capture the repo root on gitsh startup and save in the `env`

I'm more than happy to work on this should it make sense, but wanted to get a quick sanity check before going further.

georgebrock:
I think this makes a lot of sense, and I'd welcome this feature.

Another thing to consider might be a magic variable—something like `$_root`—that would evaluate to the root of the repository.

mike-burns:
Re gitsh startup: it's possible to `:cd` into another repo, and I consider that a feature.

(And I do think a bare `:cd` should go to the repo root -- that makes sense to me.)

christoomey:
I think my approach will be to detect the root dynamically at command run time with the `rev-parse` option.

If the user has `:cd`ed out of any repo, should that error or be a noop?

mike-burns:
They might have `:cd`'ed into another directory in order to run `init`.

christoomey:
Sorry, my question was worded poorly. I was referring to the case where someone had `:cd`ed out of any repo and then ran `:cd` with no argument. Should that error, or just noop? I lean towards noop, but could be convinced of providing and error message in that case.

mike-burns:
I'd be happy with noop. I'll leave the decision up to George, though.

georgebrock:
I think it should fail, in that it should return `false` behind the scenes so that `:cd && foo` wouldn't ever run `foo` outside of a Git repo. And if that's the case, it should probably output an error.

Does that seem reasonable?

mike-burns:
Good thinking.

christoomey:
[...]

PR #290, opened 2017-04-08
Repo root helpers

georgebrock:
A couple of little helpers for finding a repository's root directory:

- A `$_root` magic variable which evaluates to the repository's root.
- An update to the `:cd` command, so that it changes to the repository root when invoked without a path. Thanks to @christoomey for suggesting this one in #242.

### DIFF
--- a/spec/integration/magic_variables_spec.rb
+++ b/spec/integration/magic_variables_spec.rb
@@ -77,6 +77,21 @@ describe 'Magic variables' do
     end
   end
 
+  context '$_root' do
+    it 'evaluates to the root path of the repository' do
+      GitshRunner.interactive do |gitsh|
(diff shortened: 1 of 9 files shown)


spec/integration/magic_variables_spec.rb
```diff
@@ -77,6 +77,19 @@
     end
   end
 
+  context '$_root' do
+    it 'evaluates to the root path of the repository' do
+      GitshRunner.interactive do |gitsh|
+        root_dir = Dir.pwd
+        gitsh.type('init')
+        gitsh.type(':echo $_root')
```
mike-burns: Given this setup, this is the same as the working directory. I'd love for this test to start gitsh from a directory where the parent dir has a`.git` instead.

PR #292, opened 2017-04-22
Non-zero exit status for bad scripts.

georgebrock:
When gitsh is invoked in a non-interactive manner, a parse error should result in a non-zero exit status.

To make sure this happens, parse error handling has been delegated to the input strategies by the interpreter. In an interactive session, we output an error message and continue the session. In a non-interactive session, we raise a `Gitsh::ParseError`, which will bubble up to the gitsh binary and
terminate the session with an exit status of 1.

Since the input strategy for files is used in a few internal places (in the interactive input strategy, for loading `.gitshrc`, and in the `:source` command), those call sites have been modified to handle the possibility of a `Gitsh::ParseError`.

### DIFF
--- a/lib/gitsh/input_strategies/interactive.rb
+++ b/lib/gitsh/input_strategies/interactive.rb
@@ -48,6 +48,11 @@ module Gitsh
         retry
       end
 
+
+      def handle_parse_error(message)
+        env.puts_error("gitsh: #{message}")
+      end
+
       private
 
       attr_reader :history, :line_editor, :env, :terminal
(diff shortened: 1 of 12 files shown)


lib/gitsh/input_strategies/interactive.rb
```diff
@@ -48,6 +48,11 @@ def read_command
         retry
       end
 
+
+      def handle_parse_error(message)
+        env.puts_error("gitsh: #{message}")
```
mike-burns: Why do we not raise here? Is it because raising breaks the REPL?

mike-burns: Aha, you answered that in the PR description. Seems good.

PR #293, opened 2017-04-22
Support for multi-line commands.

georgebrock:
Line breaks are supported in similar places to sh(1). Line breaks can be escaped anywhere by ending a line with a `\` character. Unescaped line breaks are also supported:

- after logical operators (`&&` and `||`),
- within strings,
- between commands wrapped in parentheses, and
- between commands in subshells.

The `Lexer` has been expanded to insert a `MISSING` token in all situations where the input is known to be incomplete (i.e. when the input ends with an escape character, or the input ends in any of the places where an unescaped line break can be used).

After invoking the `Lexer`, the `Interpreter` checks the token stream for `MISSING` tokens. If it finds any it requests another line of input from the current input strategy, appends it to the current input, and tries again.

A new `EOL` token has been introduced to represent line breaks between commands. In the `Parser` it's treated exactly like the `SEMICOLON` token.

Lexical analysis of comments needed to be improved to allow for comments at the end of lines in a multi-line command. For example, the following input is valid:

    (:echo 1 # comment
    :echo 2)

It is semantically equivalent to:

    (:echo 1; :echo 2)

To support this, the `Lexer` will now:

- ignore whitespace before a comment's initial `#` character. This prevents extraneous `SPACE` tokens from being produced. A trailing `SPACE` token in a single line command is fine, but it can cause problems in a multi-line command.

- pop the `:comment` state without consuming the newline character at the end of a comment, allowing the default parsing rules to handle the newline, and produce an `EOL` token.

### DIFF
--- a/lib/gitsh/input_strategies/interactive.rb
+++ b/lib/gitsh/input_strategies/interactive.rb
@@ -13,6 +13,7 @@ module Gitsh
   module InputStrategies
     class Interactive
       BLANK_LINE_REGEX = /^\s*$/
+      CONTINUATION_PROMPT = '> '.freeze
 
       def initialize(opts)
         @line_editor = opts.fetch(:line_editor) do
@@ -48,6 +49,19 @@ module Gitsh
         retry
       end
 
+      def read_continuation
+        input = begin
+          line_editor.readline(CONTINUATION_PROMPT, true)
+        rescue Interrupt
+          nil
+        end
+
+        if input.nil?
+          env.print "\n"
+        end
+
+        input
+      end
 
       def handle_parse_error(message)
         env.puts_error("gitsh: #{message}")
--- a/lib/gitsh/parser.rb
+++ b/lib/gitsh/parser.rb
@@ -19,6 +19,7 @@ module Gitsh
       '!' => Gitsh::Commands::ShellCommand,
     }.freeze
 
+    left :EOL
     left :SEMICOLON
     left :OR
     left :AND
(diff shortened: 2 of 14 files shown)


lib/gitsh/input_strategies/interactive.rb
```diff
@@ -48,6 +49,19 @@ def read_command
         retry
       end
 
+      def read_continuation
+        input = begin
+          line_editor.readline(CONTINUATION_PROMPT, true)
```
georgebrock: Passing `true` here results in each line of the command having it's own entry in the history. GNU Bash combines multi-line commands into a single history entry. That's probably a worthwhile enhancement, but maybe it should be a separate PR.

mike-burns: Could improve the code documentation with weird syntax like:

```ruby
line_editor.readline(CONTINUATION_PROMPT, history_entry_per_line=true)
```

lib/gitsh/parser.rb
```diff
@@ -66,7 +68,9 @@ class Parser < RLTK::Parser
 
     production(:word, 'WORD+') { |words| words.inject(:+) }
 
-    production(:subshell, 'SUBSHELL_START .program SUBSHELL_END') { |p| p }
+    production(:subshell) do
+      clause('SUBSHELL_START .program SUBSHELL_END') { |p| p }
+    end
```
georgebrock: This is an accidental change. I must have messed up a rebase somewhere.

lib/gitsh/input_strategies/interactive.rb
```diff
@@ -13,6 +13,7 @@ module Gitsh
   module InputStrategies
     class Interactive
       BLANK_LINE_REGEX = /^\s*$/
+      CONTINUATION_PROMPT = '> '.freeze
```
sharplet: Can this make use of the `PS2` environment variable? Is that standard?

georgebrock: We don't use `PS1` for the main prompt (there's a git-config(1) variable called `gitsh.prompt` instead). If we were going to support customisation, I think we should follow that pattern instead.

sharplet: Works for me.

Issue #25, opened 2013-11-25
Support the same tab completion options as zsh


pbrisbin:
Completion in general is very lacking compared to ZSH, but I'm not sure we can anything about that.

lode:
I have a similar experience coming from [git-sh](https://github.com/rtomayko/git-sh). It feels git-sh is smarter about the context of the command.

I personally don't feel comfortable enough with ruby to start implementation, but would love to help getting this further. Would it be a good start to list some specific situations? Or is the current references to zsh and git-sh enough? (Is this a good issue btw, or is there another more generic issue already?)

mike-burns:
Specifics would be great, pull requests with failing tests would be even better, and pull requests with working code is most ideal.

See https://github.com/thoughtbot/gitsh/blob/master/CONTRIBUTING.md for details on how to get the dev environment running locally. Not too much Ruby knowledge is required for this (you'll need RubyGems and Bundler: `apt-get install rubygems bundler`). Make sure this runs and passes:

    make check

Edit `spec/units/completer_spec.rb` and copy one of the existing file path tests. Something like:

```ruby
it 'completes partial paths' do
  with_a_temporary_home_directory do
    completer = build_completer(input: 'add ')
    Dir.chdir do
      FileUtils.mkdir_p(%w(abc def ghi jkl mno))
    end

    expect(completer.call('a/d/g/j/m')).to include 'abc/def/ghi/jkl/mno'
  end
end
```

Then run the test according to CONTRIBUTING.md:

    env TESTS="spec/units/completer_spec" make -e check

If you want to tackle the Ruby after that, look in `lib/gitsh/completer.rb`, especially the `PathCompleter` class (line 102-ish). You'll need to modify the `#collection` method.

lode:
Thanks for the push into actual coding :) I tried ruby a while back and noticed it costs me a lots of time for small things, so I'll save it for later. Or, at least I'll start with some specific things I'd like to see different. Maybe later or maybe someone else can pick some up.

Some auto completion (via tab) wishes:
- `push ` would give `push origin ` (ending space)
- `merge ori` would give `merge origin/` (ending slash)
- `stash pop ` would give `stash pop stash@{` (further auto completion lists the available numbers or even the output of `stash list`)
- in general auto complete branch names instead of paths where a branch name can be expected

Often the extra space or slash at the end makes all the difference between fluently continue typing or having to type redundant stuff.

mike-burns:
[...]

PR #296, opened 2017-04-30
Replace tab completion generation system

georgebrock:
This commit replaces the `Gitsh::Completer` with a more sophisticated system. For an overview of how the new system works, see `lib/gitsh/tab_completion/README.md`, which is added by this commit.

The new system has two benefits over the old:

1. It makes use of the `Gitsh::Lexer` to split up the command, instead of naively splitting the input on spaces. This means we're able to tab complete commands that don't appear at the beginning of the line, e.g. when entering `add . && com<tab>` or `:echo $(con<tab>`.

2. It uses a non-deterministic finite automaton to understand the context of the completion, instead of just distinguishing between commands and arguments. While the NFA in this commit doesn't do any more than the old completer did, it lays the ground work for much more sophisticated context-aware completions in the future.

---

Pair: @tabfugnic


georgebrock:
This fixes #261, and takes us a big step closer to #25, #80, and #220.

lib/gitsh/tab_completion/context.rb
```diff
@@ -0,0 +1,65 @@
...
+      def prior_words
+        words[0...-1]
+      end
+
+      private
+
+      attr_reader :input
+
+      def words
+        @_words ||= tokens.reverse_each.inject(['']) do |words, token|
```
mike-burns: How about:

```ruby
def words
  combine_words(split_by_spaces(last_command))
end

def last_command
  tokens.reverse_each.take_while do |token|
    [:AND, :OR, :SEMICOLON, :LEFT_PAREN, :SUBSHELL_START, :EOL].exclude?(token.type)
  end.reverse
end

# split_by_spaces : [token] -> [[token]]
def split_by_spaces
  chunk { |token| token.type == :SPACE }.
    inject([]) do |result, (is_space, tokens)|
      if is_space
        result
      else
        result + [tokens]
      end
    end
end

# combine_words : [[word tokens]] -> [string]
def combine_words
  map do |tokens|
    tokens.select { |token| token.type == :WORD }.map(&:value).join
  end
end
```

[From pairing with @georgebrock @tabfugnic .]

lib/gitsh/tab_completion/automaton.rb
```diff
@@ -0,0 +1,79 @@
...
+      def match(tokens)
+        tokens.inject(start_states) do |current_states, token|
+          current_states.map { |state| state.follow(token) }.inject(Set.new, :|)
+        end
+      end
+
+      def completions(context, token)
+        match(context).
+          flat_map { |state| state.completions(token) }.
+          select { |option| option.start_with?(token) }
```
mike-burns: How about moving this to `BaseMatcher`, to make this automaton definition more clear?

lib/gitsh/tab_completion/context.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      attr_reader :input
+
+      def words
+        combine_words(split_by_spaces(last_command(tokens)))
+      end
+
+      def combine_words(token_groups)
+        token_groups.map do |tokens|
+          tokens.select { |token| token.type == :WORD }.map(&:value).join
```
mike-burns: Could turn these three into an `#inject`:

```ruby
tokens.inject([]) do |result, token|
  if token.type == :WORD
    "#{result}#{token.value}"
  else
    result
  end
end
```

lib/gitsh/tab_completion/matchers/path_matcher.rb
```diff
@@ -0,0 +1,34 @@
...
+            File.expand_path(token) + '/'
+          else
+            File.expand_path(token)
+          end
+        end
+
+        def paths(prefix)
+          Dir["#{prefix}*"].map do |path|
+            if File.directory?(path)
+              path + '/'
```
mike-burns: Is this too much abstraction?

```ruby
def normalize_path(token)
  conditionally_slash(File.expand_path(token)) { token.end_with?('/') }
end

def paths(prefix)
  Dir["#{prefix}*"].map do |path|
    conditionally_slash(path) { File.directory?(path) }
  end
end

def conditionally_slash(string, &predicate)
  if predicate.call
    string + '/'
  else
    string
  end
end
```

spec/units/tab_completion/matchers/text_matcher_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(matcher.completions('commit')).to eq ['commit']
+    end
+  end
+
+  describe '#eql?' do
+    it 'returns true for text matchers with the same text' do
+      x1 = described_class.new('x')
+      x2 = described_class.new('x')
+
+      expect(x1.eql?(x2)).to be true
```
mike-burns: Any reason to write it like this instead of `expect(x1).to eql(x2)`?

mike-burns:
BTW, I had some code golf ideas, but it LGTM at any point.

lib/gitsh/tab_completion/matchers/path_matcher.rb
georgebrock: I think it is a bit too much: if I saw this in practice, I'd need to read the `conditionally_slash` method to understand what was going on, so I'm going to leave it inline.

PR #298, opened 2017-05-15
Expand historyFile path

xduugu:
Expand the configured history file path, so it is possible to use a "~"
for the user's home directory instead of an absolute path.

Before, a path like "~/.local/share/gitsh_history" led to the following
error:

gitsh: Error: No such file or directory @ rb_sysopen - ~/.local/share/gitsh_history

### DIFF
--- a/lib/gitsh/history.rb
+++ b/lib/gitsh/history.rb
@@ -32,7 +32,7 @@ module Gitsh
     end
 
     def history_file_path
-      env.fetch('gitsh.historyFile') { DEFAULT_HISTORY_FILE }
+      File.expand_path(env.fetch('gitsh.historyFile') { DEFAULT_HISTORY_FILE })
     end
 
     def history_size


georgebrock:
Thanks for this!

Could you add a test for the new behaviour too? The [`persistent_history_spec`][1] might be a good place to start, and we have a [`#with_a_temporary_home_directory`][2] method that will probably be useful too.

[1]: https://github.com/thoughtbot/gitsh/blob/master/spec/integration/persistent_history_spec.rb
[2]: https://github.com/thoughtbot/gitsh/blob/master/spec/support/file_system.rb#L26-L32

georgebrock:
I added a test, and merged this into master as 1ed14c7 retaining your authorship.

Thanks!

xduugu:
Thanks for adding the test @georgebrock . I haven't found the time to have a look, so I'm glad you already merged it.

georgebrock:
@xduugu No problem. Thanks for the patch!

PR #299, opened 2017-05-21
Support pipelines.

georgebrock:
Two commands combined with the pipe character (`|`) will be run in parallel with the standard output of the first command connected to the standard input of the second command via an `IO.pipe`.

### DIFF
--- a/lib/gitsh/commands/pipeline.rb
+++ b/lib/gitsh/commands/pipeline.rb
@@ -1,0 +1,50 @@
+require 'gitsh/pipeline_environment'
+
+module Gitsh
+  module Commands
+    class Pipeline
+      def initialize(left, right)
+        @left = left
+        @right = right
+      end
+
+      def execute(env)
+        left_env, right_env = PipelineEnvironment.build_pair(env)
+        threads = [start_left_thread(left_env), start_right_thread(right_env)]
+        wait_for_threads(threads)
+        threads.map(&:value).all?
+      end
+
+      private
+
+      attr_reader :left, :right, :threads
+
+      def start_left_thread(env)
+        Thread.new { execute_left(env) }
+      end
+
+      def start_right_thread(env)
+        Thread.new { execute_right(env) }
+      end
+
+      def execute_left(left_env)
+        left.execute(left_env)
+      ensure
+        left_env.output_stream.close
+      end
+
+      def execute_right(right_env)
+        right.execute(right_env)
+      ensure
+        right_env.input_stream.close
+      end
+
+      def wait_for_threads(threads)
+        threads.map(&:join)
+      rescue Interrupt
+        threads.each { |thread| thread.raise(Interrupt) }
+        retry
+      end
+    end
+  end
+end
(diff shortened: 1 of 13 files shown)


georgebrock:
The first commit on this branch works on macOS, but not on Linux (e.g. on CI).

The difference seems to be how sh(1) handles SIGINT when run with `-c`. On Ubuntu Linux, which is the distribution CI uses and the one I've done manual testing on, `/bin/sh` appears to be dash(1). I can't reproduce the problem with FreeBSD's sh(1) or macOS's (which is GNU bash running in a compatibility mode).

To experiment, I wrote a little Ruby script that does nothing until it gets a SIGINT:

```ruby
# test.rb
begin
  sleep
rescue Interrupt
  puts 'SIGINT!'
end
```

When I spawn a new process that runs this script directly, I can easily send it an interrupt:

```
irb(main):001:0> p = Process.spawn('ruby2.0', 'test.rb')
=> 8594
irb(main):002:0> Process.kill('INT', p)
=> 1
SIGINT!
irb(main):003:0> Process.wait(p)
=> 8594
```

However, when I spawn a new process that runs the script in a shell, I can't interrupt it:

```
irb(main):004:0> p = Process.spawn('sh', '-c', 'ruby2.0 test.rb')
=> 8661
irb(main):005:0> Process.kill('INT', p)
=> 1
irb(main):006:0> Process.wait(p)
^CSIGINT!
IRB::Abort: abort then interrupt!
        from (irb):6:in `call'
        from (irb):6:in `wait'
        from (irb):6
        from /usr/bin/irb2.0:12:in `<main>'
```

Note that in this case `SIGINT!` is only output when I press <kbd><kbd>ctrl</kbd><kbd>c</kbd></kbd>.

The only vaguely useful reference to this I've found so far is a line in some [Docker documentation][1]:

> … will be started as a subcommand of `/bin/sh -c`, which does not pass signals.

[1]: https://docs.docker.com/engine/reference/builder/#entrypoint

sharplet:
Here’s another possibly relevant snippet of documentation from [OpenBSD’s `sh(1)` man page](https://man.openbsd.org/sh#SIGINT_2):

> <dl>
> <dt>SIGINT</dt>
> <dd>If a shell is interactive and in command line editing mode, editing is terminated on the current line and the command being edited is not entered into command history. Otherwise the signal is caught but no action is taken.</dd>
> </dl>

Would it make sense for `gitsh` to catch SIGINT and manually send SIGKILL to child shell processes in a pipeline?

sharplet:
Ok, I had a play around with this and I think I got it to work: https://gist.github.com/sharplet/5dc8bd4239e09f68e8f009dbfe48f9ff#file-subshell-swift-L23.

This sets up a signal handler for SIGINT that calls `kill(0, SIGTERM)`, terminating the shell and its children. I tried sending SIGTERM to just the pid of the shell, but it doesn’t also terminate child processes.

Using `0` as the first argument to `kill(2)` sends the signal to all process in the same process group. I’m not sure if that’s too blunt an instrument, but if `gitsh` isn’t running any other child processes, then terminating everything that’s currently running seems reasonable to me.

PR #300, opened 2017-05-28
Remove Ruby warnings

teoljungberg:
This PR is broken down into the commits that quells the following
warnings:

- character class has duplicated range: /X/.
  
  Where `/X/` is the regex.

- instance variable `@X` not initialized

  Where `@X` is the instance variable.

- assigned but unused variable - X

  Where `X` is the unused variable.

- ambiguous first argument; put parentheses or a space even after `/'
  operator

- character class has duplicated range: /X/

  Where `/X/` is the regex.
  `\s+` matches the output `[\n\r\s]+` does.

I leave it up to the reviewer to pick and choose the warnings that feel
applicable to cherry-pick in, that is why they are broken out into
small and concise commits.

### DIFF
--- a/lib/gitsh/arguments/subshell.rb
+++ b/lib/gitsh/arguments/subshell.rb
@@ -26,7 +26,7 @@ module Gitsh
       def strip_whitespace(output)
         output.
           sub(%r{\r?\n\Z}, '').
-          gsub(%r{[\n\r\s]+}, ' ')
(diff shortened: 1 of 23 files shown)


teoljungberg:
Also happy PR number 300!

georgebrock:
This is brilliant, thanks Teo.

PR #302, opened 2017-06-16
Improve variable lexing.

georgebrock:
Prior to this commit, an incomplete variable reference (a line ending with
an unescaped `$` character, or unclosed `${`) would produce a lexing error.
After this commit it produces a parse error instead.

We want to start using the lexer for understanding tab completion context,
which means we need to lex incomplete input. Having all legal-but-incomplete
input cause parse errors instead of lexing errors will make this much
easier.

Note that it's still possible to produce lexing errors with invalid input,
but it's more reasonable for the tab completion system to just ignore that
and not provide any completions.

A new `EXPECTED` token type was used instead of the existing `MISSING` token
type, because `MISSING` is used to determine that the user can enter another
line to complete the command.

---

**Notes for reviewers:**

- This was motivated by feedback from @mike-burns and @tabfugnic on #296, specifically the `Gitsh::TabCompletion::Context` class had some very complex code to work around incomplete input that caused lexing errors.

- I'm not wild about the distinction between `EXPECTED` and `MISSING`, maybe `MISSING` and `INCOMPLETE` would be better names.

### DIFF
--- a/lib/gitsh/interpreter.rb
+++ b/lib/gitsh/interpreter.rb
@@ -44,7 +44,7 @@ module Gitsh
     end
 
     def incomplete_command?(tokens)
-      tokens.reverse_each.detect { |token| token.type == :MISSING }
+      tokens.reverse_each.detect { |token| token.type == :INCOMPLETE }
     end
 
     def build_multi_line_command(previous_lines, new_line)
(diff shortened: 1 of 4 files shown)


mike-burns:
Code LGTM. Renaming discussed IRL:

`:MISSING` -> `:INCOMPLETE`
`:EXPECTED` -> `:MISSING`

PR #304, opened 2017-06-17
Support tab completion for variables.

georgebrock:
This commit introduces a `Gitsh::TabCompletion::VariableCompleter` object. This is completely separate from the NFA-based completion we use for most things, because:

- we immediately know we need to use variable completion based on lexical analysis of the input without having to walk through the NFA's states, and

- the context in which the variable appears doesn't inform the completions we present the user with.

The `Gitsh::TabCompletion::Facade` object gains the responsibility of deciding which type of completion to invoke, based on information from the `Gitsh::TabCompletion::Context` object.

### DIFF
--- a/lib/gitsh/environment.rb
+++ b/lib/gitsh/environment.rb
@@ -50,6 +50,14 @@ module Gitsh
       raise Gitsh::UnsetVariableError, "Variable '#{key}' is not set"
     end
 
+    def available_variables
+      (
+        magic_variables.available_variables +
+        variables.keys +
+        repo.available_config_variables
+      ).uniq.sort
+    end
+
     def config_variables
       Hash[variables.select { |key, value| key.to_s.include?('.') }]
     end
--- a/lib/gitsh/git_repository.rb
+++ b/lib/gitsh/git_repository.rb
@@ -67,6 +67,11 @@ module Gitsh
       end
     end
 
+    def available_config_variables
+      modern_git_available_config_variables ||
+        old_git_available_config_variables
+    end
+
     def config_color(name, default)
       git_output(
         "config --get-color #{Shellwords.escape(name)} #{Shellwords.escape(default)}"
@@ -115,6 +120,21 @@ module Gitsh
       sha = git_output('rev-parse HEAD')
       unless sha.empty?
         "#{sha[0,7]}..."
+      end
+    end
+
+    def modern_git_available_config_variables
+      command = git_command("config --list --name-only")
+      out, _, status = Open3.capture3(command)
+
+      if status.success?
+        out.lines.map { |line| line.chomp.to_sym }
+      end
+    end
+
+    def old_git_available_config_variables
+      git_output('config --list').lines.map do |line|
+        line.split('=').first.to_sym
       end
     end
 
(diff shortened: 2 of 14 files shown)


georgebrock:
Travis CI is using a very old version of Git (1.8.5.6) which doesn't support the `git config --list --name-only` command used by this PR.

However, the `--name-only` argument was introduced in version 2.6, and versions back to 2.4 are still officially supported.

The `--name-only` release notes indicate that it was introduced because parsing `git config --list` output is hard when there are multi-line config values. That makes me think we should probably try `--name-only` and fall back to attempting to parse `--list`.

lib/gitsh/git_repository.rb
```diff
@@ -118,6 +123,21 @@ def abbreviated_sha
...
+      out, _, status = Open3.capture3(command)
+
+      if status.success?
+        out.chomp.lines.map { |line| line.chomp.to_sym }
+      end
+    end
+
+    def old_git_available_config_variables
+      git_output('config --list').lines.map do |line|
+        line.split('=').first.to_sym
```
sharplet: This could use `line.upto('=').to_sym` to avoid scanning the entire line.

lib/gitsh/git_repository.rb
```diff
@@ -118,6 +123,21 @@ def abbreviated_sha
       end
     end
 
+    def modern_git_available_config_variables
+      command = git_command("config --list --name-only")
+      out, _, status = Open3.capture3(command)
+
+      if status.success?
+        out.chomp.lines.map { |line| line.chomp.to_sym }
```
sharplet: Is the first `#chomp` redundant given that each line is individually chomped?

lib/gitsh/environment.rb
```diff
@@ -50,6 +50,14 @@ def fetch(key, force_default_git_command = false, &block)
       raise Gitsh::UnsetVariableError, "Variable '#{key}' is not set"
     end
 
+    def available_variables
+      (
+        magic_variables.available_variables +
+        variables.keys +
+        repo.available_config_variables
+      ).sort.uniq
```
mike-burns: For the most micro of micro optimizations, you could `uniq.sort` -- this will cause sort to have a smaller list to work through.

lib/gitsh/git_repository.rb
georgebrock: Sadly, that's not what `String#upto` does (e.g. `a.upto('d')` produces an enumerator over `['a', 'b', 'c', 'd']`). Your version seems more useful though.

I could use `split('=', 2).first`, that should avoid scanning the whole thing.

lib/gitsh/git_repository.rb
georgebrock: Excellent point.

lib/gitsh/environment.rb
georgebrock: We'll save so much time over the next 50 years!

lib/gitsh/git_repository.rb
sharplet: Hmm...given that `#split` still returns an array in that case, it's unclear whether that's improving things much. The best alternative I could come up with was

```
line.each_char.take_while { |c| c != "-" }.join
```

But given that config lines aren't usually very long (I assume), it's probably not a big deal either way.

georgebrock: I think I'll stick with `split.first`.

PR #305, opened 2017-06-23
Tab completion NFA state graph visualization tool.

georgebrock:
The tab completion system that was introduced in 1b673e5 uses a
Non-deterministic Finite Automaton (NFA) to understand the user's input and
select relevant completions. The NFA has a directed graph of states.

While the current NFA's state graph is small, the intention is to replace it
with a much larger and more complex graph.

This commit introduces a visualization tool, which outputs a representation
of the NFA's state graph in Graphviz's dot language.

---

Current output:

Tab completion graph

### DIFF
--- a/lib/gitsh/tab_completion/README.md
+++ b/lib/gitsh/tab_completion/README.md
@@ -102,6 +102,18 @@ then hit <kbd>tab</kbd>. In this case, we'd pass `add` to the automaton and move
 from state `0` to state `4`. From state `4` the next thing we expect is a path,
 so we'd offer file paths from the current directory as completions.
 
+### Building the state graph
+
+The state graph is built by the `Gitsh::TabCompletion::AutomatonFactory` class.
+
+To help debug the state graph, the repository includes a visualisation tool.
+Running `bin/tcviz` will output the graph in [Graphviz's dot language][1].
+Running this through Graphviz's dot(1) program produces an image file, e.g.
+
+    ./bin/tcviz | dot -Tsvg -o tab_completion.svg
+
+[1]: http://www.graphviz.org/content/dot-language
+
 ## Major components
 
 - `Gitsh::TabCompletion::Facade` provides a single interface to the rest of
--- a/lib/gitsh/tab_completion/visualization.rb
+++ b/lib/gitsh/tab_completion/visualization.rb
@@ -1,0 +1,83 @@
+require 'set'
+require 'gitsh/tab_completion/automaton'
+require 'gitsh/tab_completion/matchers/text_matcher'
+
+module Gitsh
+  module TabCompletion
+    class Visualization
+      def initialize(automaton)
+        @automaton = automaton
+      end
+
+      def to_dot
+        %Q(digraph TabCompletion {
+          ranksep=3;
+          nodesep=3;
+          #{visitor.node_labels.join("\n  ")}
+          #{visitor.transitions.join("\n  ")}
+        })
+      end
+
+      def summary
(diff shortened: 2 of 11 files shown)


lib/gitsh/tab_completion/README.md
```diff
@@ -102,6 +102,21 @@ then hit <kbd>tab</kbd>. In this case, we'd pass `add` to the automaton and move
 from state `0` to state `4`. From state `4` the next thing we expect is a path,
 so we'd offer file paths from the current directory as completions.
 
+### Building the state graph
+
+The state graph is built by the `Gitsh::TabCompletion::AutomatonFactory` class.
+The initial implementation is small and fixed, but is expected to grow in size
+and complexity as we support more subtle forms of tab completion.
```
mike-burns: I'm not excited about this sentence/I worry about it staying around longer than this graph's simplicity.

lib/gitsh/tab_completion/README.md
```diff
@@ -102,6 +102,21 @@ then hit <kbd>tab</kbd>. In this case, we'd pass `add` to the automaton and move
...
+
+To help debug the state graph, the repository includes a visualisation tool.
+Running `bin/tcviz` will load the configuration files, and output the graph in
```
mike-burns: No comma.

lib/gitsh/tab_completion/visualization.rb
```diff
@@ -0,0 +1,97 @@
...
+      attr_reader :start_state
+
+      def node_labels
+        @node_labels ||= states.map do |state|
+          '%s [ label=%s ];' % [state_name(state), state.name.inspect]
+        end
+      end
+
+      def transitions
+        @transitions ||= states.inject([]) do |acc, state|
```
mike-burns: This is a `#flat_map` right?

georgebrock:
> I kinda expected to find `#to_dot` methods on `State`, matchers, `Automaton`, etc.

You make a good point. I originally wrote this as a throw-away script (there were a lot of `#send` calls, and it even used `ObjectSpace` to find all the `State` instances). It was surprisingly useful in debugging the `nfa-dsl-omg` branch, so I decided to keep it around, but didn't refactor as far as I should have.

I'll take another pass at it.

lib/gitsh/tab_completion/README.md
georgebrock: Good point. Also, it doesn't load any configuration files (yet—that's on the larger branch I extracted this smaller piece from).

lib/gitsh/tab_completion/README.md
georgebrock: I'll drop it.

georgebrock:
@mike-burns I decided I still wanted to keep all of the dot formatting stuff in the same object, especially because it's only useful for debugging. To avoid exposing the internals of the automaton and its states, I used the visitor pattern. Maybe this is better?

lib/gitsh/tab_completion/visualization.rb
```diff
@@ -0,0 +1,83 @@
...
+        def visit_free_transition(start_state, end_state)
+          transitions << '%s -> %s;' % [
+            state_name(start_state),
+            state_name(end_state),
+          ]
+        end
+
+        private
+
+        def state_name(state)
```
mike-burns: Maybe `state_key` or `state_label` or `state_identifier` are more appropriate names here.

georgebrock: I like `state_identifier` (`label` has meaning in the dot language)

PR #307, opened 2017-06-28
Quell ruby warnings for unused variables

teoljungberg:
This commit quells warnings along the likes of:

```
warning: assigned but unused variable - X
```

Where X is the variable in question.

This alongside https://github.com/thoughtbot/gitsh/pull/306 makes gitsh
warning free, and ensures that it stays warning free.


georgebrock:
Rebased, and reopened as #308 (retaining your authorship).

Thanks for these, Teo!

PR #311, opened 2017-07-03
Add * and ? operators to git_completions(5)

georgebrock:
This PR adds support for the `*` operator (zero or more times), and the `?` operator (zero or one times) to the gitsh tab completion DSL.

It also adds filtering for duplicate options, otherwise the more complex graphs enabled by the DSL changes can cause problems with path completion (see commit message for more details).

---

In order to motivate this change with a failing test, this PR starts by updating the tab completion configuration file to read:

    $command ($revision|$path|$remote)* --? $path+

The resulting NFA state graph looks like this:

tab_completion

### DIFF
--- a/lib/gitsh/tab_completion/dsl/maybe_operation_factory.rb
+++ b/lib/gitsh/tab_completion/dsl/maybe_operation_factory.rb
@@ -1,0 +1,19 @@
+module Gitsh
+  module TabCompletion
+    module DSL
+      class MaybeOperationFactory
+        attr_reader :child
+
+        def initialize(child)
+          @child = child
+        end
+
+        def build(start_state, options = {})
+          end_state = child.build(start_state, options)
+          start_state.add_free_transition(end_state)
+          end_state
+        end
+      end
+    end
+  end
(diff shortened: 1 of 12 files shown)


lib/gitsh/tab_completion/dsl/maybe_operation_factory.rb
```diff
@@ -0,0 +1,19 @@
+module Gitsh
+  module TabCompletion
+    module DSL
+      class MaybeOperationFactory
+        attr_reader :child
```
mike-burns: Why does this expose `#child`? I don't see it used in any of the implementation.

mike-burns:
BTW, I like your DSL. These factory objects are quite easy to read.

lib/gitsh/tab_completion/dsl/maybe_operation_factory.rb
georgebrock: I exposed them for tests, which isn't great, but the parser specs would be significantly more complex if I didn't.

PR #312, opened 2017-07-07
Tab completion DSL: Options support

georgebrock:
This PR further expands the tab completion DSL with the ability to define options like `-f` or `--force`.

- Rules can contain an `$opt` variable, which indicates where options may appear. This will match on anything beginning with `-` or `--`, and complete to any known options.
- Rules can list known options, and the arguments expected by those options.

### Examples

The config file in the repo is still pretty simple, but after this PR rules like this become possible:

```
commit $opt* --? $path+
  --all
  --fixup $revision
  --file $path
  --untracked-files (no|normal|all)?
```

Producing this state graph:

tab_completion

---

Visualisation of `etc/completions` after this PR:

tab_completion

### DIFF
--- a/lib/gitsh/tab_completion/dsl/option_transition_factory.rb
+++ b/lib/gitsh/tab_completion/dsl/option_transition_factory.rb
@@ -1,0 +1,51 @@
+require 'gitsh/tab_completion/automaton'
+require 'gitsh/tab_completion/matchers/unknown_option_matcher'
+
+module Gitsh
+  module TabCompletion
+    module DSL
+      class OptionTransitionFactory
+        def build(start_state, options = {})
+          @start_state = start_state
+          @options = options
+
+          invoke
+        end
+
+        private
+
+        attr_reader :start_state, :options
+
+        def invoke
+          add_transitions_for_known_options
+          add_transitions_for_unknown_options
+          end_state
+        end
+
+        def add_transitions_for_known_options
+          known_options_factory.build(
+            start_state,
+            options.merge(end_state: end_state),
+          )
+        end
+
+        def add_transitions_for_unknown_options
+          start_state.add_transition(
+            Matchers::UnknownOptionMatcher.new,
+            end_state,
+          )
+        end
+
+        def end_state
+          @end_state ||= options.fetch(:end_state) do
+            Automaton::State.new('option')
+          end
+        end
+
+        def known_options_factory
+          options.fetch(:known_options_factory)
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


lib/gitsh/tab_completion/dsl/option_transition_factory.rb
```diff
@@ -0,0 +1,60 @@
+require 'gitsh/tab_completion/automaton'
+require 'gitsh/tab_completion/matchers/unknown_option_matcher'
+
+module Gitsh
+  module TabCompletion
+    module DSL
+      class OptionTransitionFactory
+        def build(start_state, options = {})
+          Invocation.new(start_state, options).call
```
mike-burns: Is this better than e.g.

```ruby
def build(start_state, options = {})
  @start_state = start_state
  @options = options
  invoke
end

private

def invoke
  add_transitions_for_known_options
  add_transitions_for_unknown_options
  end_state
end

# ...
```

?

lib/gitsh/tab_completion/dsl/rule_factory.rb
```diff
@@ -2,14 +2,15 @@ module Gitsh
...
 
-        def initialize(root)
+        def initialize(root, options)
           @root = root
+          @options = options
         end
 
         def build(start_state)
-          root.build(start_state)
+          root.build(start_state, context: self)
```
mike-burns: Should this make use of `options`?

lib/gitsh/tab_completion/dsl/option_transition_factory.rb
georgebrock: Perhaps not. I'm not a big fan of setting ivars outside of `#initialize` (unless they're memoising a lazily calculated value), but the inner class makes this harder to read.

lib/gitsh/tab_completion/dsl/rule_factory.rb
georgebrock: I suppose we could pass `options: options` instead of `context: self` here—the only thing we use the `:context` value for is to call its `#options` method.

lib/gitsh/tab_completion/dsl/option_transition_factory.rb
mike-burns: I phrase it as "don't set ivars outside of the constructor", and `#build` is the constructor.

georgebrock: I think the thing I like about that pattern is the clear separation between setting up an object, and doing work. We'd lose that separation here, but the result is overall less complex than the nested class, so I'll go with it.

PR #313, opened 2017-07-12
Tab completions DSL: configuration file

georgebrock:
This tab completion configuration file includes:

- All Git commands without arguments.
- Popular Git commands, with their options and arguments.
- Internal gitsh commands, with their options and arguments.

The resulting state graph has 575 nodes and 1,627 edges, and is too large for the visualisation tool to produce anything very useful.

There are some argument types that we can't currently complete, and others that are unlikely to ever be practical to complete. To represent those this commit also introduces a `$anything` variable to the tab completion DSL. The variable's corresponding `AnythingMatcher` class will match any input, but produce no completions.

### DIFF
--- a/etc/completions
+++ b/etc/completions
@@ -1,5 +1,926 @@
-$command $opt* ($revision|$path|$remote)* --? $path+
+# This file defines tab completion options for gitsh(1).
+# The file format is described by gitsh_completions(5).
+
+
+# Common Git commands
+# ===================
+
+(add|stage) $opt* --? $path+
+  --all
+  --chmod $anything
+  --dry-run
+  --edit
+  --force
+  --ignore-errors
+  --ignore-missing
+  --ignore-removal
+  --intent-to-add
+  --interactive
+  --no-all
+  --no-ignore-removal
+  --patch
+  --refresh
+  --update
+  --verbose
+
+blame $opt* $revision? --? $path
+  --abbrev $anything
+  --contents $path
+  --date $anything
+  --encoding $anything
+  --incremental
+  --indent-heuristic
+  --line-porcelain
+  --no-indent-heuristic
+  --no-progress
+  --porcelain
+  --progress
+  --reverse $revision
+  --reverse $revision
+  --root
+  --score-debug
+  --show-email
+  --show-name
+  --show-number
+  --show-stats
+  --since $anything
+  -L $anything
+  -S $path
+
+# Listing
+branch $opt* $anything*
+  --abbrev $anything
+  --all
+  --color (always|never|auto)
+  --column $anything
+  --contains $revision
+  --ignore-case
+  --list
+  --merged $revision
+  --no-abbrev
+  --no-color
+  --no-column
+  --no-contains $revision
+  --no-merged $revision
+  --points-at $anything
+  --remotes
+  --sort $anything
+  --verbose
+
+# Creating
+branch $opt* $anything $revision?
+  --create-reflog
+  --force
+  --no-create-reflog
+  --no-track
+  --quiet
+  --set-upstream
+  --track
+
+# Modifying
+branch $opt $anything?
+  --edit-description
+  --set-upstream-to $anything
+  --unset-upstream
+  -u $anything
+
+# Renaming
+branch $opt+ $anything? $anything
+  --force
+  --move
+  -M
+  -m
+
+# Deleting
+branch $opt* $anything+
+  --delete
+  --force
+  --quiet
+  --remotes
+  -d
+  -D
+
+checkout $opt* $revision?
+  --confict (merge|diff3)
+  --detach
+  --force
+  --ignore-other-worktrees
+  --ignore-skip-worktree-bits
+  --merge
+  --no-progress
+  --no-recurse-submodules
+  --no-track
+  --orphan $anything
+  --ours
+  --patch
+  --progress
+  --quiet
+  --recurse-submodules
+  --theirs
+  --track
+  -b $anything
+  -B $anything
+
+checkout $opt* $anything? --? $path+
+
+cherry-pick $opt* $revision+
+  --allow-empty
+  --allow-empty-message
+  --edit
+  --ff
+  --gpg-sign $anything?
+  --keep-redundant-commits
+  --mainline $anything
+  --no-commit
+  --signoff
+  --strategy (resolve|recursive|octopus|ours|subtree)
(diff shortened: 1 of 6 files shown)


georgebrock:
Note to reviewers: the bulk of this change is the `etc/completions` file, but the diff is too large for GitHub to show it by default. Click to load!

etc/completions
```diff
@@ -1,5 +1,807 @@
...
+  --edit
+  --force
+  --ignore-errors
+  --ignore-missing
+  --ignore-removal
+  --intent-to-add
+  --interactive
+  --no-all
+  --no-ignore-removal
+  --refresh
```
mike-burns: Missing `--patch`.

etc/completions
```diff
@@ -1,5 +1,807 @@
-$command $opt* ($revision|$path|$remote)* --? $path+
+(add|stage) $opt* --? $path+
+  --all
+  --dry-run
```
mike-burns: Missing `--chmod`.

spec/units/tab_completion/matchers/anything_matcher_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      matcher = described_class.new(double(:env))
+
+      expect(matcher.completions('foo')).to eq([])
+      expect(matcher.completions('--all')).to eq([])
+      expect(matcher.completions('')).to eq([])
+    end
+  end
+
+  describe '#eql?' do
+    it 'returns true when given another instance of the same class' do
```
rickreilly: Do you prefer conditional modifiers to `context`s?

etc/completions
```diff
@@ -1,5 +1,807 @@
...
+  --track
+  --no-track
+  --force
+  --create-reflog
+  --quiet
+
+# Modifying
+branch $opt $anything?
+  --edit-description
+  --set-upstream-to $anything
```
mike-burns: Missing `--unset-upstream`.

etc/completions
```diff
@@ -1,5 +1,807 @@
-$command $opt* ($revision|$path|$remote)* --? $path+
+(add|stage) $opt* --? $path+
```
rickreilly: Should there be a headlining comment of some sort at the top here?

etc/completions
```diff
@@ -1,5 +1,807 @@
...
+  --show-stats
+  --abbrev $anything
+  --contents $path
+  --date $anything
+  --encoding $anything
+  --reverse $revision
+  --since $anything
+  -S $path
+
+# Listing
```
rickreilly: `# Branch listing`? 

Also, is this a YAML file? If so, would using indentation help convey the hierarchy more clearly?

spec/units/tab_completion/matchers/anything_matcher_spec.rb
georgebrock: I do in cases like this, where there would be one `it` per `context`, and the combined description is relatively short.

etc/completions
```diff
@@ -1,5 +1,807 @@
...
+  --no-all
+  --no-ignore-removal
+  --refresh
+  --update
+  --verbose
+
+blame $opt* $revision? --? $path
+  --incremental
+  --line-porcelain
+  --no-progress
```
rickreilly: Should all these options be alphabetized?

etc/completions
georgebrock: Good call. It would probably be helpful to not only have a section header comment (like I do below), but also reference the manual page where users can find out about the file format.

etc/completions
georgebrock: Nope, it's a custom file format described in this manual page: https://github.com/thoughtbot/gitsh/blob/1f7c9615384bf7661cb5557918350c46ebc99b46/man/man5/gitsh_completions.5.in

It's a bit like regular expressions, but the atoms are command arguments not characters.

rickreilly:
Is there any way to integration test this? And/or any way to check this against any sort of comprehensive list of Git commands/flags?

georgebrock:
> View changes I started comparing against my git man pages but found things in here that aren't documented in the man page (e.g. `branch --delete`) and things in the man page that aren't in here.

Thanks for reviewing this in that much detail!

> So I'm going to pause and ask: which git version, and where did you pull these from?

I pulled them from the manual pages, using Git version 2.9.1. Interestingly, some of the things you mention aren't referenced in the manual pages I have, but appear to be valid options (e.g. `add --chmod` isn't mentioned in my version of git-add(1), but gives the error `error: option `chmod' requires a value`)

georgebrock:
> Is there any way to integration test this? And/or any way to check this against any sort of comprehensive list of Git commands/flags?

The existing integration specs in [`spec/integration/tab_completion_spec.rb`](https://github.com/thoughtbot/gitsh/blob/1f7c9615384bf7661cb5557918350c46ebc99b46/spec/integration/tab_completion_spec.rb) will load and use this config. While I didn't change them for this PR, they are now completing based on these rules.

It would probably be beneficial to add a test there for tab completing an option. This is the first version of the config file that's defined commands with options.

georgebrock:
> So I'm going to pause and ask: which git version, and where did you pull these from?

I remember thinking at some point: maybe in the future we'll need to provide different tab completion definitions for different versions of Git. Maybe the future is now?

mike-burns:
I'm on

```
~% git --version
git version 2.11.0
```

How do you want to handle versioning?

georgebrock:
> How do you want to handle versioning?

Three things occur to me:

1. Don't worry about it just yet: make sure this matches the latest release, and open an issue to support completions for older versions, or newer versions when they're released.
2. Define a completion file per Git version, e.g. I would load `tab_completion_2.9` and you would load `tab_completion_2.11`.
3. Define some baseline completions file that works all the way back to some old-but-still-supported-version, and then define additional files for each subsequent version that add support for their new features, e.g. you'd load `base`, `2.9`, `2.10`, and `2.11`; I'd stop at `base` and `2.9`. That would cut down on duplication, but it wouldn't give us any way to drop support for removed features.

mike-burns:
I think (1) is right for now.

etc/completions
georgebrock: On the `# Branch listing` thing—I think the `branch` part is clear from the context.

georgebrock:
I've upgraded to Git 2.13.3, and will make the completions consistent with that version's manual pages.

georgebrock:
@rickreilly @mike-burns Finished updating for Git 2.13, and ready for another look.

georgebrock:
We're up to 673 nodes and 1905 edges after these updates!

Issue #323, opened 2017-12-06
"Which git" prompt placeholder


benknoble:
My gitsh prompt contains the string `git` at the end of the prompt, both to remind me of my context and to let me know that everything is getting a git prefix.

```
project (feature-branch @) git> 
```
I get this with `%D (%c%b %#%w) git> `.

I have `hub` installed, and use it frequently, but typically *only* to interact with github (again, this is for context purposes). However, occasionally I want to fire up gitsh with hub (`gitsh --git "$(which hub)"`). Most often I do this in a split tmux pane via a hotkey.

I would like to be able to see the "git binary" being used in my prompt.

Would it be possible for gitsh to support a prompt placeholder that resolves to the currently-used git binary?

Something like
```
%g        basename of current git binary
%G        full name of current git binary
```

Then the prompt `%D (%c%b %#%w) %g> ` would become
- `project (feature-branch @) git> ` when run normally
- `project (feature-branch @) hub> ` when run with `hub`
- &c.

This extends to `gh` and other git providers (??) as well.

georgebrock:
Great suggestion, thanks @benknoble.

This probably overlaps a bit with #231 too.

benknoble:
Actually, in keeping with the current gitsh prompt strings, I think it makes sense to reverse the meaning of `%g` and `%G` mentioned above.

benknoble:
Furthermore, were I to try to implement this, would it make sense to strip `usr/bin/env` from the "full path" to binary" and `env` from the basename?

Without stripping, we get things like

```
project (feature-branch @) /usr/bin/env git> 
project (feature-branch @) env git> 
```

when run normally, which I'm not a fan of, but I could see a use-case of something like `gitsh --git "/usr/bin/env hub"` (weird as it seems) where knowing that `env` was used might be important.

PR #324, opened 2017-12-19
Implement git_command placeholder in prompt

benknoble:
See #323 

This required a fairly minimal change to the pattern-matching substitution when generating the prompt string. I also updated the relevant section of the man page.

The tests fail without my code and pass with it, though it should be noted that this was my first experience with ruby and spec.

### DIFF
--- a/spec/units/prompter_spec.rb
+++ b/spec/units/prompter_spec.rb
@@ -166,6 +166,26 @@ describe Gitsh::Prompter do
           "#{File.basename(Dir.getwd.sub(/\A#{Dir.home}/, '~'))} "
         )
       end
+
+      it 'replaces %g with the absolute path of the current git binary' do
+        env = env_double(
+          format: '%g',
+          git_command: '/usr/local/bin/my-custom-git',
+        )
+        prompter = Gitsh::Prompter.new(env: env)
+
+        expect(prompter.prompt).to eq '/usr/local/bin/my-custom-git '
+      end
+
+      it 'replaces %G with the basename of the current git binary' do
+        env = env_double(
+          format: '%G',
+          git_command: '/usr/local/bin/my-custom-git',
+        )
+        prompter = Gitsh::Prompter.new(env: env)
+
+        expect(prompter.prompt).to eq 'my-custom-git '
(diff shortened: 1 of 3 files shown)


spec/units/prompter_spec.rb
```diff
@@ -166,6 +166,22 @@
...
+
+      it 'replaces %g with the absolute path of the current git binary' do
+        env = env_double(format: '%g')
+        prompter = Gitsh::Prompter.new(env: env)
+
+        expect(prompter.prompt).to eq "#{env.git_command} "
+      end
+
+      it 'replaces %G with the basename of the current git binary' do
+        env = env_double(format: '%G')
```
georgebrock: What do you think about stubbing `git_command` here, so that we can make the assertion more explicit? e.g. if we stubbed it as `/usr/local/bin/my-custom-git`, we could verify the output was `my-custom-git`.

benknoble: More than happy to, as I rather like the idea of a more explicit test. Not sure how to stub it—would I do that with the `env_double` below or somewhere else?

georgebrock: In `#env_double`, I think you could drop `allow(env).to receive(:git_command).and_return(…)` and instead add `git_command: …` to the `default_attrs` Hash. That would allow you to override the value of `git_command` by calling `env_double(git_command: 'some custom value')`

benknoble: Cool. Having never worked with ruby before, it's remarkably expressive.

benknoble:
Sqaushed!

benknoble:
Is development on Gitsh planned to continue? I realize it seems to have stagnated since last January (and fully admit that you, as the maintainer, are likely *very* busy).

I'm mostly just wondering if/when we can expect to see a return the project, as it has proved immensely useful for me.

georgebrock:
@benknoble You're totally right that development has been very slow lately, but I still use gitsh every day as do other people at thoughtbot, and I'm sure we'd pick up anything urgent that came up (compatibility issues with a new version of macOS, say).

PR #326, opened 2018-03-06
Remove extra spaces in Markdown-rendered code

tysongach:
Each of the code blocks in the readme (with the exception of the first, which uses HTML instead of Markdown) have two extra spaces at the front of each line. This removes those. I believe GitHub changed its Markdown parser somewhat recently; perhaps this was an artifact of that.

The rendered diff looks like this:

screen shot 2018-03-05 at 20 40 58


georgebrock:
Thanks, Tyson.

PR #331, opened 2018-11-12
Custom completions file

### LINKED ISSUE #315: Support a user-specific completions file
As well as the global completions file in `$prefix/etc`, we should support a user-specific completions file in the user's home directory. This would allow users to define completions for their custom Git commands and aliases.

georgebrock:
This PR adds support for a `$HOME/.gitsh_completions` file, so that users can add tab completion rules for their custom Git commands and aliases.

Resolves #315

### DIFF
--- a/lib/gitsh/tab_completion/automaton_factory.rb
+++ b/lib/gitsh/tab_completion/automaton_factory.rb
@@ -13,7 +13,10 @@ module Gitsh
       end
 
       def build
-        load_config(File.join(env.config_directory, 'completions'))
+        start_state = Automaton::State.new('start')
+        config_paths.each do |path|
+          DSL.load(path, start_state, env)
+        end
         Automaton.new(start_state)
       end
 
@@ -21,12 +24,11 @@ module Gitsh
 
       attr_reader :env
 
-      def load_config(path)
-        DSL.load(path, start_state, env)
-      end
-
-      def start_state
-        @start_state ||= Automaton::State.new('start')
+      def config_paths
+        [
+          File.join(env.config_directory, 'completions'),
+          File.join(ENV.fetch('HOME', '/'), '.gitsh_completions'),
+        ]
       end
     end
   end
--- a/spec/integration/tab_completion_spec.rb
+++ b/spec/integration/tab_completion_spec.rb
@@ -155,4 +155,19 @@ describe 'Completing things with tab' do
       expect(gitsh).to output(/\AThird\nSecond\n\Z/)
     end
   end
+
+  it 'completes with custom rules' do
+    with_a_temporary_home_directory do |home|
+      write_file("#{home}/.gitsh_completions", 'recho $revision')
+      GitshRunner.interactive do |gitsh|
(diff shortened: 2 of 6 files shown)


spec/support/file_system.rb
```diff
@@ -52,7 +52,7 @@ def switch_home_directory(&block)
       ENV['HOME'] = path
       write_file("#{path}/.inputrc", DEFAULT_READLINE_CONFIG)
       write_file("#{path}/.gitconfig", DEFAULT_GIT_CONFIG)
-      block.call
+      block.call(path)
```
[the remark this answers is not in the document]
mike-burns: Hey Rubocop,

https://bugs.ruby-lang.org/issues/14045
https://bugs.ruby-lang.org/issues/14330

spec/integration/tab_completion_spec.rb
```diff
@@ -155,4 +155,19 @@
...
     end
   end
+
+  it 'completes with custom rules' do
+    with_a_temporary_home_directory do |home|
+      write_file("#{home}/.gitsh_completions", 'recho $revision')
+      GitshRunner.interactive do |gitsh|
+        gitsh.type('init')
+        gitsh.type('commit --allow-empty -m First')
+        gitsh.type('config --local alias.recho "!echo"')
```
mike-burns: I'm picturing this pronounced as Scooby-Doo saying "echo".

lib/gitsh/tab_completion/automaton_factory.rb
```diff
@@ -13,20 +13,22 @@ def initialize(env)
...
-      def load_config(path)
-        DSL.load(path, start_state, env)
-      end
-
-      def start_state
-        @start_state ||= Automaton::State.new('start')
+      def config_paths
+        [
+          File.join(env.config_directory, 'completions'),
+          File.join(ENV['HOME'], '.gitsh_completions'),
```
mike-burns: So there's an unlikely but possible chance that `$HOME` is unset -- for example, if I set `gitsh` as my login shell, `$HOME` may be unset.

That would mean that it reads the `.gitsh_completions` from `.`, I think. That feels a bit unpredictable -- the exploit there is that if I run gitsh from an untrusted directory, someone could drop a malformed gitsh_completions file in there.

OTOH, it might be handy -- you can get project-specific completions depending on where you start running gitsh. In which case `:cd` should also do this.

Either way, how about being more exact here:

```suggestion
          File.join(ENV.fetch('HOME', '/'), '.gitsh_completions'),
```

or

```suggestion
          File.join(ENV.fetch('HOME', '.'), '.gitsh_completions'),
```

georgebrock: Good catch, thanks. I think falling back to `/.gitsh_completions` makes sense when `$HOME` isn't set—that would also be consistent with what would happen with the `.gitshrc` file.

Project-specific completions are an interesting idea, but I'm not sure I see a use case that would only make sense for a specific project.

Issue #325, opened 2018-02-22
Expand tilde (~) and environment variables in paths


sharplet:
I'd like some more flexibility in the kinds of external commands I can run in gitsh, for example, when I want to copy a file from another repo:

```
!cp ~/src/some-repo/some-file ./
cp: ~/src/some-repo/some-file: No such file or directory
```

The next thing I tried was this:

```
!cp $HOME/src/some-repo/some-file ./
gitsh: Variable 'HOME' is not set
```

So I suppose this is actually two different shell-related feature requests: ~ expansion, and access to environment variables.

georgebrock:
Since #321 covers `~`, I'm updating this issue to focus on just environment variables.

Issue #321, opened 2017-11-14
Support tilde "~" expansion for external commands


sharplet:
I just tried the following command and was surprised when it didn't work:

```
my-app my-branch! !mv activity.json ~/Desktop/                                                                                                                                                                                                                       
mv: rename activity.json to ~/Desktop/: No such file or directory   
```

Would it be possible to support tilde expansion for external commands?

Issue #303, opened 2017-06-17
Resizing terminal window while in a pager confuses the next pager


georgebrock:
1. Open gitsh in a small terminal emulator window
2. Run a command that starts a pager
3. Increase the window size
4. Close the pager
5. Run a command that starts a pager
6. The new pager will be initialised at the original smaller window size

This is similar to the issue discussed in https://github.com/thoughtbot/gitsh/issues/144#issuecomment-69459773 and fixed in #221.

We might be able to fix this by updating environment variables on SIGWINCH. The less(1) manual page says this about the `COLUMNS` environment variable:

> Sets the number of columns on the screen.  Takes precedence over the number of columns specified by the TERM variable.  (But if you have a windowing system which supports `TIOCGWINSZ` or `WIOCGETD`, the window system's idea of the screen size takes precedence over the `LINES` and `COLUMNS` environment variables.)

Issue #145, opened 2014-03-19
Wildcards don't seem to work correctly


leifg:
In gitsh (version 0.5, without any special git binary) I get the following error:

```
diff Gemfile*
```

results in

```
fatal: ambiguous argument 'Gemfile*': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

However `git diff Gemfile*` on the shell works correctly

Any ideas?

georgebrock:
Another benefit to this: implementing our own globbing would mean we could stop passing shell commands through `/bin/sh -c`, which in turn would unblock pipelines (#299)

georgebrock:
With the new tab completion system we can now generate a context-aware list of possible arguments, which we could use as the basis of custom globbing.

It could work something like this:

1. User enters a command with a glob pattern in one or more arguments, e.g. `show-branch */master my-feature-branch`
2. We see there's a glob in an argument, and ask the tab completion system for possible completions at that point in the command, e.g. pass `"show-branch "` to the tab completion system, and get back options like `["origin/master", "upstream/master", "--all", "v0.1", …]`
3. We filter the options based on the pattern, e.g. we reduce the options to `["origin/master", "upstream/master"]`
4. We expand the argument to the available options, e.g. our example command becomes `show-branch origin/master upstream/master my-feature-branch`

sharplet:
Here’s another example where the lack of wildcard expansion has unexpected results:

```
$ git rm -r --cached */xcuserdata
rm 'CallForSupport.xcodeproj/xcuserdata/adsharp.xcuserdatad/xcschemes/xcschememanagement.plist’

# vs.

$ git rm -r --cached “*/xcuserdata”
fatal: pathspec '*/xcuserdata' did not match any files
```

In `gitsh`, because wildcards aren’t expanded, running `rm -r --cached */xcuserdata` is equivalent to the second invocation, where quoting prevents the shell from expanding the wildcard.

sharplet:
Further observations:

- When `zsh(1)` is unable to successfully expand a glob pattern, it fails with the error `zsh: no matches found: */xcuserdat`
- When `bash(1)` is unable to successfully expand a glob pattern, it invokes the command with the literal string as an argument, resulting in `fatal: pathspec '*/xcuserdat' did not match any files`

PR #335, opened 2018-11-21
Add GOALS file, to define 1.0.

georgebrock:
What's left to do before we can call this 1.0?

### DIFF
--- a/GOALS
+++ b/GOALS
@@ -1,0 +1,7 @@
+gitsh hits 1.0 when:
+
+[x] it provides a useful interactive shell environment.
+[x] it runs scripts.
+[ ] it has the kinds of features expected by advanced shell users.
+[ ] its tab completion system is at least as good as Git's bash integration.
+[ ] it has a smooth install process on popular versions of macOS, BSD and GNU/Linux.


georgebrock:
This was inspired by a Slack conversation about versioning, and https://github.com/thoughtbot/terrapin/blob/master/GOALS

A lot of these things are described in more detail in GitHub issues, so maybe a GitHub Milestone would be a better place to put this?

GOALS
```diff
@@ -0,0 +1,7 @@
+gitsh hits 1.0 when:
+
+[x] it provides a useful interactive shell environment.
+[x] it runs scripts.
+[ ] it has the kinds of features expected by advanced shell users.
```
tabfugnic: This seems a bit vague. Should this have examples?

GOALS
```diff
@@ -0,0 +1,7 @@
...
+[ ] its tab completion system is at least as good as Git's bash integration.
+[ ] it has a smooth install process on popular versions of macOS, BSD and GNU/Linux.
```
tabfugnic: Happy to help with this!

GOALS
```diff
@@ -0,0 +1,7 @@
+gitsh hits 1.0 when:
+
+[x] it provides a useful interactive shell environment.
+[x] it runs scripts.
```
tabfugnic: While this is done, should this be a bit more descriptive?

mike-burns:
Yeah or maybe a GH Issue Project?

In a Slack thread I listed these advantages:

- give everyone an indication of what it means for them to use a sub-1.0 project.
- give me a clear indication of when I've hit 1.0, so I won't run into perfectionism stopping me.
- help me get to 1.0 ASAP, so we can get normal semver going.

GOALS
georgebrock: I was thinking of these features: #321, #325, #145, #100, #67, and maybe also #303, #41, #35

Not sure if it's redundant to list them out here and in GitHub issues? I could also imagine someone suggesting a very reasonable feature that fits in this list and seems like it should be in 1.0, but maybe I should resist that kind of feature creep.

georgebrock:
After the conversations here about repeating issues in a GOALS file, I set up a couple of issue milestones: https://github.com/thoughtbot/gitsh/milestones

I think these do the same job, although maybe the should be referenced in the contributing guide.

mike-burns:
Oh that's great. LGTM. I guess you'd close this?

georgebrock:
Yep, closing this in favour of GitHub milestones.

PR #337, opened 2018-12-14
Git 2 20 tab complete

georgebrock:
This PR changes how gitsh builds a full list of available commands for tab
completion.

This list was originally built by parsing the output of `git help -a`, which
was based on the approach taken by the tab completion scripts distributed
with Git for various general-purpose shells.

In Git version 2.20, the output of `git help -a` changed in ways that
weren't compatible with our existing parsing code. Passing `--no-verbose`
continues to provide the old output format. Meanwhile, a new `--list-cmds`
argument was added, which outputs commands as a simple list, but is also
described in the git(1) manual page as being experimental and subject to
change.

To make sure we continue supporting a range of Git versions, and we won't
have major problems if the `--list-cmds` argument is removed, this PR
introduces the following strategy for getting a full list of Git commands:

1. Try `git --list-cmds`
2. Try `git help -a --no-verbose`
3. Try `git help -a`
4. Fail with an empty array

Since this is much more complex to the previous approach, we first extract
an object to handle this one responsibility.

### DIFF
--- a/lib/gitsh/git_command_list.rb
+++ b/lib/gitsh/git_command_list.rb
@@ -1,0 +1,29 @@
+module Gitsh
+  class GitCommandList
+    def initialize(env)
+      @env = env
+    end
+
+    def to_a
+      git_output('help -a').
+        lines.
+        select { |line| line =~ /^  [a-z]/ }.
+        map { |line| line.split(/\s+/) }.
+        flatten.
+        reject { |cmd| cmd.empty? || cmd =~ /--/ }.
+        sort
+    end
+
+    private
+
+    attr_accessor :env
+
+    def git_output(command)
+      Open3.capture3(git_command(command)).first.chomp
+    end
+
+    def git_command(sub_command)
+      "#{env.git_command} #{sub_command}"
+    end
+  end
+end
--- a/lib/gitsh/git_command_list.rb
+++ b/lib/gitsh/git_command_list.rb
@@ -5,8 +5,33 @@ module Gitsh
     end
 
     def to_a
-      git_output('help -a').
-        lines.
+      try_using(commands_from_list_cmds) do
+        try_using(commands_from_help('help -a --no-verbose')) do
+          try_using(commands_from_help('help -a'))
+        end
+      end
+    end
+
+    private
+
+    attr_accessor :env
+
+    def try_using(result, default: [])
+      if result && result.any?
+        result
+      elsif block_given?
+        yield
+      else
+        default
+      end
+    end
+
+    def commands_from_list_cmds
+      git_output('--list-cmds=main,nohelpers').sort
+    end
+
+    def commands_from_help(command)
+      git_output(command).
         select { |line| line =~ /^  [a-z]/ }.
         map { |line| line.split(/\s+/) }.
         flatten.
@@ -14,12 +39,14 @@ module Gitsh
         sort
     end
 
-    private
-
-    attr_accessor :env
-
     def git_output(command)
-      Open3.capture3(git_command(command)).first.chomp
+      output, _, status = Open3.capture3(git_command(command))
+
+      if status.success?
+        output.chomp.lines.map(&:chomp)
+      else
+        []
+      end
     end
 
     def git_command(sub_command)
--- a/lib/gitsh/git_repository.rb
+++ b/lib/gitsh/git_repository.rb
@@ -1,6 +1,7 @@
 require 'open3'
 require 'shellwords'
 require 'gitsh/git_repository/status'
+require 'gitsh/git_command_list'
 
 module Gitsh
   class GitRepository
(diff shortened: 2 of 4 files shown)


lib/gitsh/git_command_list.rb
```diff
@@ -5,7 +5,38 @@ def initialize(env)
...
+
+    attr_accessor :env
+
+    def commands_from_list_cmds
+      output = git_output('--list-cmds=main,nohelpers')
+
+      if output
+        output.lines.map(&:chomp).sort
+      else
+        false
```
mike-burns: How about returning `[]` in the failure case, and then instead of `||` you make a helper method that calls the next result if the result is `empty?`?

lib/gitsh/git_command_list.rb
```diff
@@ -0,0 +1,62 @@
...
+      output, _, status = Open3.capture3(git_command(command))
+
+      if status.success?
+        output.chomp
+      else
+        false
+      end
+    end
+
+    def git_command(sub_command)
```
mike-burns: I'd like to state that I acknowledge that this is duplicated, have mild concerns about this, and am overall OK with it.

lib/gitsh/git_command_list.rb
georgebrock: To make that lazy (which I think it should be, because each of these runs an external command) it would probably need to use some nested blocks. I guess that would be fine though. Something like:

```ruby
try_using(commands_from_list_cmds) do
  try_using(commands_from_help('help -a --no-verbose')) do
    try_using(commands_from_help('help -a'), default: [])
  end
end
```

lib/gitsh/git_command_list.rb
georgebrock: Yep, that did happen. It also reminded me how much we pass a `Gitsh::Environment` instance around, which I'd love to get away from eventually.

lib/gitsh/git_command_list.rb
mike-burns: Yeah.

But on the up-side, this method gets simpler:

```ruby
git_output('--list-cmds=main,nohelpers')&.lines.map(&:chomp).sort || []
```

though maybe it doesn't help the other two...

sharplet: I don’t know which version of Ruby you’re targeting, but would some combination of `&.` and `yield_self` be appropriate?

```ruby
def commands_from_list_cmds
  git_output('--list-cmds=main,nohelpers')&.
    lines.
    map(&:chomp).
    sort
end

def commands_from_help(command)
  git_output(command)&.yield_self(:parse_help_output)
end
```

georgebrock: Currently, we support Ruby 2.0 and up (https://github.com/thoughtbot/gitsh/blob/master/configure.ac#L30-L34) but since 2.0–2.2 are now EOL'd, I'd be open to requiring 2.3.

`&.` was introduced in 2.3, but `#yield_self` is probably still out of reach in 2.5.

georgebrock: I've tried going down the `#try_using` route, combined with making sure `#git_output` always returns an array of lines (which could be empty). I think the result is much better—what do y'all think?

lib/gitsh/git_command_list.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+      end
+    end
+
+    private
+
+    attr_accessor :env
+
+    def try_using(result, default = [])
+      if result && result.any?
```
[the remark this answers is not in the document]
georgebrock: As discussed in elsewhere on this PR: we currently support Ruby 2.0+, so we don't have this operator.

lib/gitsh/git_command_list.rb
```diff
@@ -0,0 +1,56 @@
...
+          try_using(commands_from_help('help -a'))
+        end
+      end
+    end
+
+    private
+
+    attr_accessor :env
+
+    def try_using(result, default = [])
```
sharplet: What do you think about using a defaulted named parameter here? That way if a default is provided, it’s a little more self-documenting.

lib/gitsh/git_command_list.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    def commands_from_list_cmds
+      git_output('--list-cmds=main,nohelpers').sort
+    end
+
+    def commands_from_help(command)
+      git_output(command).
+        select { |line| line =~ /^  [a-z]/ }.
+        map { |line| line.split(/\s+/) }.
+        flatten.
```
mike-burns: I tried re-writing this whole block as one `#inject` -- so one iteration through the elements -- but this `#flatten` defeated me. What you have is much more readable than the `#inject` variant.

georgebrock: Thanks for trying the `#inject` route. I'll merge this as-is.

PR #341, opened 2019-01-20
Documentation tweaks

georgebrock:
- OS X has been called macOS for a while
- gitsh development requires GNU Autotools


georgebrock:
I'm not sure if a Brewfile is beneficial at this point.

There are only two Homebrew dependencies right now, and I don't think `less Brewfile; brew bundle` is going to save time over `brew install readline; brew install automake`.

PR #342, opened 2019-01-22
Change process name to gitsh by embedding Ruby in C

### LINKED ISSUE #41: The process name should be `gitsh`, not `ruby`

sharplet:
Fixes #41.

This resolves the process name issue by running a Ruby interpreter in-process instead of via `ruby(1)`. The implementation is inspired by the article [Running Ruby in C](https://silverhammermba.github.io/emberb/embed/) and [Ruby's own `main` function](https://github.com/ruby/ruby/blob/bdbc8a8f1214b601ca409f66f2fe00ac06f2e99a/main.c). Instead of invoking `ruby(1)` as a subprocess, we use `ruby.h` and `libruby` to perform the same function directly.

In order to test this, I hard-coded the paths to `ruby.h`, `ruby/config.h` and `libruby` on my machine. The full set of compiler flags, including the platform-specific location of `ruby/config.h`, are available via `pkg-config(1)`. I'm not very familiar with Autotools, so I could use some help figuring out how to correctly include those flags at build time.

### DIFF
--- a/src/gitsh.c
+++ b/src/gitsh.c
@@ -1,12 +1,27 @@
 #include <err.h>
+#include <ruby.h>
 #include <sysexits.h>
-#include <unistd.h>
 
 int
 main(int argc, char *argv[])
 {
-    if (execv(GITSH_RB_PATH, argv) == -1) {
-        err(EX_SOFTWARE, GITSH_RB_PATH);
+    int ruby_argc;
+    char **ruby_argv;
+
+    ruby_argc = argc + 2;
+    ruby_argv = calloc(ruby_argc + 1, sizeof(char *));
+
+    if (!ruby_argv) {
+        err(EX_OSERR, GITSH_RB_PATH);
     }
-    return EX_OK;
+
+    ruby_argv[0] = argv[0];
+    ruby_argv[1] = "--disable-gems";
+    ruby_argv[2] = GITSH_RB_PATH;
+    memcpy(&ruby_argv[3], &argv[1], argc * sizeof(char *));
+
+    ruby_sysinit(&argc, &argv);
+    ruby_init();
+
+    return ruby_run_node(ruby_options(ruby_argc, ruby_argv));
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,15 @@ rvm:
   - "2.4"
   - "2.5"
   - "2.6"
-install: bundle install && ./autogen.sh && git config --global user.name 'Travis' && git config --global user.email 'travis@example.com'
-script: RUBY=$(which ruby) ./configure && make && bundle exec rspec
+install:
+  - bundle install
+  - ./autogen.sh
+  - git config --global user.name 'Travis'
+  - git config --global user.email 'travis@example.com'
+  - export LD_LIBRARY_PATH="${MY_RUBY_HOME}/lib"
+script:
+  - RUBY=$(which ruby) ./configure --prefix=$PWD/tmp
+  - make
+  - make install
+  - echo :help | ./tmp/bin/gitsh
+  - bundle exec rspec
--- a/Makefile.am
+++ b/Makefile.am
@@ -3,6 +3,7 @@ EXTRA_DIST = \
 	LICENSE \
 	README.md \
 	INSTALL \
+	configure.rb \
 	vendor/vendorize \
 	ext/gitsh/extconf.rb \
 	ext/gitsh/src/line_editor.c \
@@ -13,13 +14,16 @@ EXTRA_DIST = \
 
 bin_PROGRAMS = gitsh
 gitsh_SOURCES = src/gitsh.c
-AM_CPPFLAGS = -DGITSH_RB_PATH="\"$(rubydir)/$(PACKAGE).rb\""
+gitsh_CPPFLAGS = $(RUBY_CFLAGS) \
+	-DGITSH_RB_PATH="\"$(rubydir)/$(PACKAGE).rb\""
+gitsh_LDFLAGS = $(RUBY_LIBS)
+LDADD = $(RUBY_LDADD)
 
 man_MANS = man/man1/gitsh.1 man/man5/gitsh_completions.5
 nobase_dist_pkgdata_DATA = $(vendorfiles)
 nobase_dist_ruby_DATA = $(libfiles)
 pkgrubylib_DATA = lib/gitsh/version.rb
-ruby_SCRIPTS = src/gitsh.rb
+ruby_DATA = src/gitsh.rb
 dist_pkgsysconf_DATA = etc/completions
 
 CLEANFILES = src/gitsh.rb lib/gitsh/version.rb
@@ -33,7 +37,6 @@ substitute_values = sed \
 
 src/gitsh.rb: Makefile src/gitsh.rb.in
 	$(substitute_values) $(srcdir)/$@.in > $@
-	chmod +x $@
 
 lib/gitsh/version.rb: Makefile lib/gitsh/version.rb.in
 	$(substitute_values) $(srcdir)/$@.in > $@
--- a/configure.ac
+++ b/configure.ac
@@ -33,6 +33,13 @@ AX_PROG_RUBY_VERSION(
   AC_MSG_ERROR(Ruby 2.3.0 or later is required to install gitsh)
 )
 
+RUBY_CFLAGS=$($RUBY ./configure.rb '-I$(rubyarchhdrdir) -I$(rubyhdrdir)')
+RUBY_LIBS=$($RUBY ./configure.rb '-L$($(libdirname)) $(LIBRUBYARG) $(LIBS)')
+RUBY_LDADD=$($RUBY ./configure.rb '$($(libdirname))/$(LIBRUBY)')
+AC_SUBST([RUBY_CFLAGS])
+AC_SUBST([RUBY_LIBS])
+AC_SUBST([RUBY_LDADD])
+
 VENDOR_DIRECTORY="vendor/gems"
 
 test -d $VENDOR_DIRECTORY || mkdir -p $VENDOR_DIRECTORY
--- a/configure.rb
+++ b/configure.rb
@@ -1,0 +1,18 @@
+require 'rbconfig'
+
+VAR_PATTERN = /\$\((?<name>[a-z_]+)\)/i
+
+REPLACEMENTS = Hash[RbConfig::MAKEFILE_CONFIG.map do |key, value|
+  ["$(#{key})", value]
+end]
+
+def expand_config(path)
+  if path =~ VAR_PATTERN
+    new_path = path.gsub(VAR_PATTERN, REPLACEMENTS)
+    expand_config(new_path)
+  else
+    path
+  end
+end
+
+puts expand_config(ARGV[0])
--- a/src/gitsh.rb.in
+++ b/src/gitsh.rb.in
@@ -1,5 +1,3 @@
-#!@RUBY@ --disable-gems
-
 $LOAD_PATH.unshift('@rubylibdir@')
 
 require '@gemsetuppath@'


sharplet:
Another thought: with this change, would it make sense to unset the executable bit on `src/gitsh.rb` and remove the shebang line? https://github.com/thoughtbot/gitsh/blob/da37a35974878b39a69da2405c15792a7306dbe2/src/gitsh.rb.in#L1

georgebrock:
Thanks for this!

We may be able to use pkg-config from autoconf via the [`PKG_CHECK_MODULES` macro](https://autotools.io/pkgconfig/pkg_check_modules.html), but I'm not sure if we should: it's not always installed, and even if it is we can't guarantee there will be a `ruby.pc` file somewhere useful, e.g. on macOS I see a `/usr/local/lib/pkgconfig/ruby.pc` file for Homebrew's Ruby, but no `/usr/lib/pkgconfig/ruby.pc` for the default system Ruby (it's possible this does exist in a different location, but `pkg-config --list-all | grep -i ruby` only seems to find the Homebrew version). 

If we don't use pkg-config, we can maybe handle Ruby the same way as Readline, and assume the right environment variables are set for the build system to find everything. This adds a bit of overhead for manual builds, but should either be true by default when installing from a package manager, or at least a one-time concern for the package maintainer. If we go down this path, we could maybe provide environment files to make development easier on various platforms, e.g. we could then recommend something like `source env/macos_homebrew; ./configure` for development on macOS.

Alternatively, we also have a `Gitsh::LineEditor` module which is a native extension, and configured using Ruby's [MakeMakefile (`mkmf`) library](https://ruby-doc.org/stdlib-2.5.1/libdoc/mkmf/rdoc/MakeMakefile.html). We might be able to somehow use mkmf to discover the location of `ruby.h` etc. in the configure script, and use that information to set the right environment variables for the binary to build.

@mike-burns As the OBSD package maintainer, I'm curious what you think about these approaches?

src/gitsh.c
```diff
@@ -1,12 +1,25 @@
...
 int
 main(int argc, char *argv[])
 {
-    if (execv(GITSH_RB_PATH, argv) == -1) {
-        err(EX_SOFTWARE, GITSH_RB_PATH);
+    int ruby_argc;
+    char *ruby_argv[];
+
+    ruby_argc = argc + 1;
+    ruby_argv = malloc((ruby_argc + 1) * sizeof(char *));
```
mike-burns: Can we use `calloc(3)` instead of `malloc(3)` with multiplication? That way we don't have to worry about `ruby_argc` being a really, really large number and then underflowing.

mike-burns:
I'm a fan of `pkg-config`, but it's concerning that macOS does not ship with a `ruby.pc` for its system Ruby. I wonder why. Do we know which systems we'd be losing if we went with `pkg-config`?

---


In case you want to see `pkg-config` in action:

1. https://github.com/bitptr/bytestream/blob/master/configure.ac#L7 - for `gtk+-3.0`, which I named `GTK`.
2. https://github.com/bitptr/bytestream/blob/master/Makefile.am#L5-L6 - add to the `CFLAGS` and `LDFLAGS`, but in a way such that the package builder can override them.

src/gitsh.c
sharplet: TIL, and thanks for the tip! The linked article was also a joy to read.

georgebrock:
I discovered that there _is_ a `ruby.pc` for the macOS system Ruby, it's just not in a place that pkg-config(1) is configured to look by default. On my system, it's here: `/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/pkgconfig/ruby-2.3.pc`

Since we already need to find a compatible Ruby binary in the configure script, we might be able to tap into the metadata mkmf uses to work out where to look for `ruby.pc`, e.g. `$RUBY -r rbconfig -e 'puts RbConfig::MAKEFILE_CONFIG["libdir"]'` produces `/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr` with my system Ruby

georgebrock:
Also interesting:

```
$ /usr/bin/ruby -r rbconfig -e 'puts RbConfig::MAKEFILE_CONFIG["ruby_pc"]'
ruby-2.3.pc
```

src/gitsh.c
```diff
@@ -1,12 +1,25 @@
...
 #include <sysexits.h>
-#include <unistd.h>
 
 int
 main(int argc, char *argv[])
 {
-    if (execv(GITSH_RB_PATH, argv) == -1) {
-        err(EX_SOFTWARE, GITSH_RB_PATH);
+    int ruby_argc;
+    char *ruby_argv[];
```
georgebrock: I get a compile error on this line when I try to build this locally:

```
src/gitsh.c:9:11: error: definition of variable with array type needs an
      explicit size or an initializer
    char *ruby_argv[];
          ^
1 error generated.
```

This version built successfully:

```suggestion
    char **ruby_argv;
```

sharplet: Yeah, I ran into this myself 👍 pushing untested code…

georgebrock:
I think I have a working implementation of using `ruby -r rbconfig` to work out what the `PKG_CONFIG_PATH` should be, and then using `PKG_CHECK_MODULES` in `configure.ac` to figure out the right compiler and linker flags to build this.

I say that I _think_ this is working, because on my macOS Mojave system the compiler arguments provided pkg-config don't work: `ruby.h` is missing. Some quick Web searching suggests this is a common problem and can be fixed by some `xcode-select` or manual install shenanigans. The implementation did work with rubies installed via Homebrew and rbenv.

@sharplet I've added this to the branch as db6b516 – does it work with macOS system ruby for you?

src/gitsh.c
georgebrock: I'll push a commit to fix this, to see what happens on CI

georgebrock:
I tried a new experiment in the last few commits, and set flags based on what Ruby's mkmf does to build a Makefile for a native extension. My logic is that gitsh already builds a native extension (`Gitsh::LineEditor`), so whatever flags we're using in that mkmf-generated Makefile must be correct.

It's possible that the approach is good, but the current set of flags I'm passing definitely aren't.

mike-burns:
I wonder whether either of these autoconf macros will help:

https://www.gnu.org/software/autoconf-archive/ax_ruby_ext.html#ax_ruby_ext

> Fetches the linker flags and C compiler flags for compiling and linking Ruby binary extensions.

https://www.gnu.org/software/autoconf-archive/ax_ruby_devel.html#ax_ruby_devel

> This macro checks for Ruby and tries to get the include path to `ruby.h`.

mike-burns:
@mike-burns note to self: `ruby-module(5)` has the docs you'll need to figure out how to package this. You might need to add`GEM_BIN_SUFFIX` to the PLIST entries for the compiled binary. And maybe you should already be doing this?

georgebrock:
Green test suite! 🍏 

I'm still seeing one error locally: when I build against macOS system Ruby (`/usr/bin/ruby`, which is 2.3.7p456) the build is successful, but then the resulting binary crashes with the following error:

```
gitsh: Error: $0 not initialized
```

If I remove the error handling, here's the stack trace:

```
/Users/georgebrock/gitsh-test/share/gitsh/ruby/lib/gitsh/cli.rb:13:in `initialize': $0 not initialized (RuntimeError)
        from /Users/georgebrock/gitsh-test/share/gitsh/ruby/gitsh.rb:11:in `new'
        from /Users/georgebrock/gitsh-test/share/gitsh/ruby/gitsh.rb:11:in `<main>'
```

Binaries built against Homebrew Ruby and rbenv Ruby were both fine for me locally.

georgebrock:
@mike-burns Good find! Skimming `AX_RUBY_DEVEL`, it looks like it's taking a similar direction to the last few commits here, but it looks much more comprehensive. We should try switching to that.

mike-burns:
In case you're wondering how to simply drop one of those in:

1. `mkdir m4`
2. `curl -LO http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_ruby_devel.m4`
3. `aclocal -I m4`

https://github.com/bitptr/argonaut/blob/master/autogen.sh - an example autogen script for a project that uses autoconf macros.

https://github.com/bitptr/argonaut/blob/master/configure.ac#L7 - then I use the macro.

georgebrock:
I think I'm using the macro correctly, but it's failing to link a test program in Travis CI and locally with Rubies I built with rbenv.

The linker flags it's using look like they have the same problems we had with pkg-config, i.e. they are missing a `-L` option for the directory where the Ruby shared object can be found.

I'm going to revert b4f0bd7 for now, but if anyone wants to take another look at `AX_RUBY_DEVEL` that commit, and the associated [CI build][1] are possibly a good starting point.

[1]: https://travis-ci.org/thoughtbot/gitsh/builds/484935652

mike-burns:
The issue is `RbConfig::CONFIG["LIBRUBYARG_SHARED"]` right?

This is mine (OpenBSD Jan 20):

```
~% ruby -rrbconfig -e 'puts RbConfig::CONFIG["LIBRUBYARG_SHARED"]'
-Wl,-rpath,/usr/local/lib -L/usr/local/lib -lruby24
```

It looks like CI is getting

```
-Wl,-rpath,'${ORIGIN}/../lib' -Wl,-R'${ORIGIN}/../lib' -lruby
```

Does your `ruby.pc` have a `LIBRUBYARG_SHARED` in it?

```
~% grep LIBRUBYARG_SHARED /usr/local/lib/pkgconfig/ruby-2.5.pc
LIBRUBYARG_SHARED=-Wl,-rpath,${libdir} -L${libdir} -l${RUBY_SO_NAME}
LIBRUBYARG=${LIBRUBYARG_SHARED}
Libs: ${DLDFLAGS} ${LIBRUBYARG_SHARED} ${LIBS}
```

It's not really clear to me where the `-L` comes from in `LIBRUBYARG_SHARED`. Somewhere in here https://github.com/ruby/ruby/blob/5e7d03531e109269e3aff7d9865bdf11f3308e67/configure.ac ?

georgebrock:
Here's what I see for one of my rbenv rubies:

```
$ ruby -rrbconfig -e 'puts RbConfig::CONFIG["LIBRUBYARG_SHARED"]'

```

There is a `LIBRUBYARG_SHARED` in the `.pc` file, but it's also blank:

```
$ grep LIBRUBYARG_SHARED /Users/georgebrock/.rbenv/versions/2.6.0/lib/pkgconfig/ruby-2.6.pc
LIBRUBYARG_SHARED=
Libs: ${DLDFLAGS} ${LIBRUBYARG_SHARED} ${LIBS}
```

The `DLDFLAGS` does include a `-L`, but that doesn't seem to be used by the macro:

```
$ grep -E '.?-L' /Users/georgebrock/.rbenv/versions/2.6.0/lib/pkgconfig/ruby-2.6.pc
DLDFLAGS=-L/Users/georgebrock/.rbenv/versions/2.6.0/lib  -Wl,-undefined,dynamic_lookup -Wl,-multiply_defined,suppress
$ ruby -rrbconfig -e 'puts RbConfig::CONFIG["DLDFLAGS"]'
-L/Users/georgebrock/.rbenv/versions/2.6.0/lib  -Wl,-undefined,dynamic_lookup -Wl,-multiply_defined,suppress
```

mike-burns:
Though you know, you only need `-L` if it's not in the default.

Here's OpenBSD again:

```
~% sudo ldconfig -v | grep /usr/local
~% 
```

Do you knokw what the default library search path is on macOS? Some quick searches indicate that this might be in `DYLD_FALLBACK_LIBRARY_PATH`, or perhaps documented in dyld(1).

Moreover, do you know what the error is that configure is getting?

I tried this:

```
~% cat > foo.c
#include <ruby.h>
~% clang -c `ruby -rrbconfig -e 'print RbConfig::CONFIG["LIBRUBYARG_SHARED"]'` -I `ruby -rrbconfig -e 'print RbConfig::CONFIG["rubyhdrdir"]'` -I `ruby -rrbconfig -e 'print RbConfig::CONFIG["rubyarchhdrdir"]'` foo.c -o foo.o
clang: warning: -Wl,-rpath,/usr/local/lib: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: -lruby24: 'linker' input unused [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-L/usr/local/lib' [-Wunused-command-line-argument]
~% file foo.o
foo.o: ELF 64-bit LSB relocatable, x86-64, version 1
```

I don't feel very confident about the `-I` in there (it depends on those strings having exactly one value), but it might also Just Work for you, too.

src/gitsh.c
georgebrock: Done!

georgebrock:
I made a couple more tweaks, and this is now working for me on every permutation of macOS Ruby that I've tested, which includes:

- System Ruby (ruby 2.3.7p456 on macOS 10.14.3)
- A couple of versions of Homebrew's bottled Ruby (ruby 2.5.0p0, and ruby 2.6.2p47)
- A patch version from each supported minor version installed via ruby-build and rbenv (ruby 2.6.2p47, ruby 2.5.5p157, ruby 2.4.5p335, ruby 2.3.7p456)

The changes to get all of these working were:

- Add a call to `ruby_sysinit` before `ruby_init` to prevent crashes on versions 2.4 and older.
- Add `RbConfig::MAKEFILE_CONFIG['LIBS']` to `configure.ac`'s `RUBY_LIBS`, specifically because some rubies were missing libgmp.

In order to be confident this is portable, I'd like to at least test on the platforms we usually check before a release:

- [x] macOS
- [ ] GNU/Linux (usually Debian)
- [ ] FreeBSD
- [ ] OpenBSD

georgebrock:
I tried to test Linux compatibility on CI (see afe123b) and got [failures because of a missing shared object file][1]. Needs more work to figure out if that's a GNU/Linux thing, a Travis thing, or a rvm thing.

I've reverted that change for now to see CI run cleanly on other commits on the branch.

Given the closer integration with Ruby, I think it's important to run the binary on CI so we can catch regressions early if the C API changes in future versions of Ruby.

[1]: https://travis-ci.org/thoughtbot/gitsh/jobs/515003739

mike-burns:
What am I doing wrong?

system ruby: `ruby 2.3.3p222 (2016-11-21) [x86_64-linux-gnu]`

```
checking for ruby version... ruby 2.3.3p222 (2016-11-21) [x86_64-linux-gnu]
/usr/local/lib/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- bundler (LoadError)
        from /usr/local/lib/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
        from ./vendor/vendorize:3:in `<main>'
configure: error: Vendorizing gems failed
```

setrb ruby: `ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-linux]`

```
checking for ruby version... ruby 2.3.3p222 (2016-11-21) [x86_64-linux-gnu]
Traceback (most recent call last):
        7: from ./vendor/vendorize:15:in `<main>'
        6: from ./vendor/vendorize:15:in `open'
        5: from ./vendor/vendorize:16:in `block in <main>'
        4: from /home/mike/.rubies/ruby-2.6.0/lib/ruby/2.6.0/bundler/definition.rb:237:in `specs_for'
        3: from /home/mike/.rubies/ruby-2.6.0/lib/ruby/2.6.0/bundler/definition.rb:170:in `specs'
        2: from /home/mike/.rubies/ruby-2.6.0/lib/ruby/2.6.0/bundler/spec_set.rb:85:in `materialize'
        1: from /home/mike/.rubies/ruby-2.6.0/lib/ruby/2.6.0/bundler/spec_set.rb:85:in `map!'
/home/mike/.rubies/ruby-2.6.0/lib/ruby/2.6.0/bundler/spec_set.rb:91:in `block in materialize': Could not find filigree-0.3.3 in any of the sources (Bundler::GemNotFound)
configure: error: Vendorizing gems failed
```

georgebrock:
@mike-burns Looks like for your system ruby, you need to `gem install bundler && bundle` and for your setrb ruby, you just need to `bundle`.

mike-burns:
Thanks for that, George. Has missed the obvious.

setrb ruby: `ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-linux]`

```
~/thoughtbot/gitsh% bundle                                                      
Fetching gem metadata from https://rubygems.org/........                        
Using bundler 1.17.2
Using coderay 1.1.2                                                             
Using diff-lcs 1.3
Using ffi 1.10.0                                                                
Fetching filigree 0.3.3
Installing filigree 0.3.3                                                      
Using method_source 0.9.2                                                       
Using pry 0.12.2                           
Fetching rltk 3.0.1                                                             
Installing rltk 3.0.1                                         
Using rspec-support 3.8.0                                                       
Using rspec-core 3.8.0          
Using rspec-expectations 3.8.2                                                  
Using rspec-mocks 3.8.0                   
Fetching rspec 3.8.0                                                            
Installing rspec 3.8.0
Bundle complete! 3 Gemfile dependencies, 13 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
~/thoughtbot/gitsh% ruby -v
ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-linux]
~/thoughtbot/gitsh% ./configure --prefix=/tmp/gitsh
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes                               
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... gawk                                                       
checking whether make sets $(MAKE)... yes                           
checking whether make supports nested variables... yes                        
checking for gcc... gcc                                                        
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking whether gcc understands -c and -o together... yes
checking for style of include used by make... GNU
checking dependency style of gcc... gcc3
checking for ruby... /usr/bin/ruby
checking for ruby... (cached) /usr/bin/ruby
checking for a sed that does not truncate output... /bin/sed
checking for grep that handles long lines and -e... /bin/grep
checking for ruby version... ruby 2.3.3p222 (2016-11-21) [x86_64-linux-gnu]
checking for tgetnum() in -lncurses... yes
checking for readline/readline.h... yes
checking for readline/history.h... yes
checking for readline() in -lreadline... yes
checking for rl_set_screen_size() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completion_append_character in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completion_suppress_quote in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_editing_mode in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_line_buffer in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_char_is_quoted_p in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completion_quote_character in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_getc() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_getc_function() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_filename_completion_function() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_username_completion_function() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completion_matches() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_refresh_line() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_deprep_term_function in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completer_word_break_characters in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completer_quote_characters in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_attempted_completion_over in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_library_version in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_point in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_event_hook in stdio.h,readline/readline.h,readline/history.h...yes
checking for rl_catch_sigwinch in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_catch_signals in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_pre_input_hook in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_special_prefixes in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_cleanup_after_signal() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_free_line_state() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_clear_signals() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_get_screen_size() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_vi_editing_mode() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_emacs_editing_mode() in stdio.h,readline/readline.h,readline/history.h... yes
checking for replace_history_entry() in stdio.h,readline/readline.h,readline/history.h... yes
checking for remove_history() in stdio.h,readline/readline.h,readline/history.h... yes
checking for clear_history() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_redisplay() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_insert_text() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_delete_text() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_hook_func_t* in stdio.h,readline/readline.h,readline/history.h... yes
checking for rb_obj_reveal() in ruby.h... yes
checking for rb_obj_hide() in ruby.h... yes
creating Makefile
checking that generated files are newer than configure... done
configure: creating ./config.status
config.status: creating Makefile
config.status: creating INSTALL
config.status: executing depfiles commands
~/thoughtbot/gitsh% make
gcc -DPACKAGE_NAME=\"gitsh\" -DPACKAGE_TARNAME=\"gitsh\" -DPACKAGE_VERSION=\"0.1
4\" -DPACKAGE_STRING=\"gitsh\ 0.14\" -DPACKAGE_BUGREPORT=\"hello@thoughtbot.com\
" -DPACKAGE_URL=\"\" -DPACKAGE=\"gitsh\" -DVERSION=\"0.14\" -I.  -I/usr/include/
x86_64-linux-gnu/ruby-2.3.0 -I/usr/include/ruby-2.3.0 -DGITSH_RB_PATH="\"/tmp/gi
tsh/share/gitsh/ruby/gitsh.rb\""   -g -O2 -MT src/gitsh-gitsh.o -MD -MP -MF src/
.deps/gitsh-gitsh.Tpo -c -o src/gitsh-gitsh.o `test -f 'src/gitsh.c' || echo './
'`src/gitsh.c
mv -f src/.deps/gitsh-gitsh.Tpo src/.deps/gitsh-gitsh.Po
gcc  -g -O2 -L/usr/lib -lruby-2.3 -lpthread -lgmp -ldl -lcrypt -lm   -o gitsh sr
c/gitsh-gitsh.o /usr/lib/libruby-2.3.so.2.3.0
gcc: error: /usr/lib/libruby-2.3.so.2.3.0: No such file or directory
Makefile:644: recipe for target 'gitsh' failed
make: *** [gitsh] Error 1
```

system: `ruby 2.3.3p222 (2016-11-21) [x86_64-linux-gnu]`

```
% ./configure --prefix=/tmp/gitsh
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking whether gcc understands -c and -o together... yes
checking for style of include used by make... GNU
checking dependency style of gcc... gcc3
checking for ruby... /usr/bin/ruby
checking for ruby... (cached) /usr/bin/ruby
checking for a sed that does not truncate output... /bin/sed
checking for grep that handles long lines and -e... /bin/grep
checking for ruby version... ruby 2.3.3p222 (2016-11-21) [x86_64-linux-gnu]
Unpacked gem: '/home/mike/thoughtbot/gitsh/vendor/gems/ffi-1.10.0'
Unpacked gem: '/home/mike/thoughtbot/gitsh/vendor/gems/filigree-0.3.3'
Unpacked gem: '/home/mike/thoughtbot/gitsh/vendor/gems/rltk-3.0.1'
checking for tgetnum() in -lncurses... yes
checking for readline/readline.h... yes
checking for readline/history.h... yes
checking for readline() in -lreadline... yes
checking for rl_set_screen_size() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completion_append_character in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completion_suppress_quote in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_editing_mode in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_line_buffer in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_char_is_quoted_p in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completion_quote_character in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_getc() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_getc_function() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_filename_completion_function() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_username_completion_function() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completion_matches() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_refresh_line() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_deprep_term_function in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completer_word_break_characters in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_completer_quote_characters in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_attempted_completion_over in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_library_version in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_point in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_event_hook in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_catch_sigwinch in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_catch_signals in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_pre_input_hook in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_special_prefixes in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_cleanup_after_signal() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_free_line_state() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_clear_signals() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_get_screen_size() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_vi_editing_mode() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_emacs_editing_mode() in stdio.h,readline/readline.h,readline/history.h... yes
checking for replace_history_entry() in stdio.h,readline/readline.h,readline/history.h... yes
checking for remove_history() in stdio.h,readline/readline.h,readline/history.h... yes
checking for clear_history() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_redisplay() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_insert_text() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_delete_text() in stdio.h,readline/readline.h,readline/history.h... yes
checking for rl_hook_func_t* in stdio.h,readline/readline.h,readline/history.h... yes
checking for rb_obj_reveal() in ruby.h... yes
checking for rb_obj_hide() in ruby.h... yes
creating Makefile
checking that generated files are newer than configure... done
configure: creating ./config.status
config.status: creating Makefile
config.status: creating INSTALL
config.status: executing depfiles commands
% make
gcc -DPACKAGE_NAME=\"gitsh\" -DPACKAGE_TARNAME=\"gitsh\" -DPACKAGE_VERSION=\"0.14\" -DPACKAGE_STRING=\"gitsh\ 0.14\" -DPACKAGE_BUGREPORT=\"hello@thoughtbot.com\" -DPACKAGE_URL=\"\" -DPACKAGE=\"gitsh\" -DVERSION=\"0.14\" -I.  -I/usr/include/x86_64-linux-gnu/ruby-2.3.0 -I/usr/include/ruby-2.3.0 -DGITSH_RB_PATH="\"/tmp/gitsh/share/gitsh/ruby/gitsh.rb\""   -g -O2 -MT src/gitsh-gitsh.o -MD -MP -MF src/.deps/gitsh-gitsh.Tpo -c -o src/gitsh-gitsh.o `test -f 'src/gitsh.c' || echo './'`src/gitsh.c
mv -f src/.deps/gitsh-gitsh.Tpo src/.deps/gitsh-gitsh.Po
gcc  -g -O2 -L/usr/lib -lruby-2.3 -lpthread -lgmp -ldl -lcrypt -lm   -o gitsh src/gitsh-gitsh.o /usr/lib/libruby-2.3.so.2.3.0 
gcc: error: /usr/lib/libruby-2.3.so.2.3.0: No such file or directory
Makefile:644: recipe for target 'gitsh' failed
make: *** [gitsh] Error 1
```

I have the Debian `ruby-dev` package installed. Oh, I never said this:

```
% uname -a
Linux exploitation 4.9.0-8-amd64 #1 SMP Debian 4.9.130-2 (2018-10-27) x86_64 GNU/Linux
```

Here's another thing:

```
% locate libruby
/home/mike/.rubies/ruby-2.3.0/lib/libruby-static.a
/home/mike/.rubies/ruby-2.3.1/lib/libruby-static.a
/home/mike/.rubies/ruby-2.3.8/lib/libruby-static.a
/home/mike/.rubies/ruby-2.4.0/lib/libruby-static.a
/home/mike/.rubies/ruby-2.4.1/lib/libruby-static.a
/home/mike/.rubies/ruby-2.4.2/lib/libruby-static.a
/home/mike/.rubies/ruby-2.4.3/lib/libruby-static.a
/home/mike/.rubies/ruby-2.5.0/lib/libruby-static.a
/home/mike/.rubies/ruby-2.5.1/lib/libruby-static.a
/home/mike/.rubies/ruby-2.5.3/lib/libruby-static.a
/home/mike/.rubies/ruby-2.6.0/lib/libruby-static.a
/home/mike/.rubies/ruby-2.6.1/lib/libruby-static.a
/home/mike/.rubies/ruby-2.6.2/lib/libruby-static.a
/usr/lib/x86_64-linux-gnu/libruby-2.3-static.a
/usr/lib/x86_64-linux-gnu/libruby-2.3.so
/usr/lib/x86_64-linux-gnu/libruby-2.3.so.2.3
/usr/lib/x86_64-linux-gnu/libruby-2.3.so.2.3.0
/usr/share/doc/libruby2.3
/usr/share/doc/libruby2.3/changelog.Debian.gz
/usr/share/doc/libruby2.3/changelog.gz
/usr/share/doc/libruby2.3/copyright
/usr/share/lintian/overrides/libruby2.3
/usr/share/systemtap/tapset/libruby2.3-x86_64-linux-gnu.stp
/var/cache/apt/archives/libruby2.3_2.3.3-1_amd64.deb
/var/lib/dpkg/info/libruby2.3:amd64.list
/var/lib/dpkg/info/libruby2.3:amd64.md5sums
/var/lib/dpkg/info/libruby2.3:amd64.shlibs
/var/lib/dpkg/info/libruby2.3:amd64.symbols
/var/lib/dpkg/info/libruby2.3:amd64.triggers
```

georgebrock:
@mike-burns Thanks! That's really helpful.

It looks like maybe we could resolve this by changing this line in `configure.ac`:

```
RUBY_LDADD=$($RUBY ./configure.rb '$(libdir)/$(LIBRUBY)')
```

To something like this:

```
RUBY_LDADD=$($RUBY ./configure.rb '$(archlibdir)/$(LIBRUBY)')
```

Unfortunately, that would break things on macOS! There's hopefully some other `RbConfig` variable we can use, like `$(libdirname)`, to work out the right thing to do on each system.

Can you post your `RbConfig` data somewhere? This command should do it:

```
ruby -r rbconfig -r pp -e 'pp RbConfig::MAKEFILE_CONFIG'
```

Here's the output from macOS system Ruby for comparison: https://gist.github.com/georgebrock/82d2072fd29070e8612d77a7d3f194b1

mike-burns:
This worked:

```diff
gitsh[as-process-name]& diff
diff --git a/configure.ac b/configure.ac
index 05302d9..9ef4c69 100644
--- a/configure.ac
+++ b/configure.ac
@@ -34,8 +34,8 @@ AX_PROG_RUBY_VERSION(
 )

 RUBY_CFLAGS=$($RUBY ./configure.rb '-I$(rubyarchhdrdir) -I$(rubyhdrdir)')
-RUBY_LIBS=$($RUBY ./configure.rb '-L$(libdir) $(LIBRUBYARG) $(LIBS)')
-RUBY_LDADD=$($RUBY ./configure.rb '$(libdir)/$(LIBRUBY)')
+RUBY_LIBS=$($RUBY ./configure.rb '-L$($(libdirname)) $(LIBRUBYARG) $(LIBS)')
+RUBY_LDADD=$($RUBY ./configure.rb '$($(libdirname))/$(LIBRUBY)')
 AC_SUBST([RUBY_CFLAGS])
 AC_SUBST([RUBY_LIBS])
 AC_SUBST([RUBY_LDADD])
```

georgebrock:
That works on macOS, too!

mike-burns:
Got really far in OpenBSD. It requires clang. Note to future Mike: `make CC=clang`.

configure, make, and make install work. Running it gives a surprising error:

```
$ /tmp/gitsh/bin/gitsh --version
Traceback (most recent call last):
	14: from /tmp/gitsh/share/gitsh/ruby/gitsh.rb:7:in `<main>'
	13: from /usr/local/lib/ruby/2.6/rubygems/core_ext/kernel_require.rb:54:in `require'
	12: from /usr/local/lib/ruby/2.6/rubygems/core_ext/kernel_require.rb:54:in `require'
	11: from /tmp/gitsh/share/gitsh/ruby/lib/gitsh/cli.rb:5:in `<top (required)>'
	10: from /usr/local/lib/ruby/2.6/rubygems/core_ext/kernel_require.rb:54:in `require'
	 9: from /usr/local/lib/ruby/2.6/rubygems/core_ext/kernel_require.rb:54:in `require'
	 8: from /tmp/gitsh/share/gitsh/ruby/lib/gitsh/input_strategies/interactive.rb:2:in `<top (required)>'
	 7: from /usr/local/lib/ruby/2.6/rubygems/core_ext/kernel_require.rb:54:in `require'
	 6: from /usr/local/lib/ruby/2.6/rubygems/core_ext/kernel_require.rb:54:in `require'
	 5: from /tmp/gitsh/share/gitsh/ruby/lib/gitsh/file_runner.rb:1:in `<top (required)>'
	 4: from /usr/local/lib/ruby/2.6/rubygems/core_ext/kernel_require.rb:54:in `require'
	 3: from /usr/local/lib/ruby/2.6/rubygems/core_ext/kernel_require.rb:54:in `require'
	 2: from /tmp/gitsh/share/gitsh/ruby/lib/gitsh/interpreter.rb:1:in `<top (required)>'
	 1: from /usr/local/lib/ruby/2.6/rubygems/core_ext/kernel_require.rb:54:in `require'
/usr/local/lib/ruby/2.6/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- rltk (LoadError)
```

```
$ uname -a
OpenBSD stream.my.domain 6.5 GENERIC#808 amd64
```

mike-burns:
Nevermind, I didn't bundle correctly. It works on OpenBSD!

georgebrock:
@sharplet @mike-burns Thanks so much for all the work on this PR! It feels great to be closing out a 5 year old issue.

We now have this building consistently on macOS, GNU/Linux (including CI), OpenBSD, and FreeBSD. I'm going to merge it 🎉

sharplet:
Hooray! Thanks @georgebrock and @mike-burns for getting it over the line!

PR #344, opened 2019-01-24
Update supported Ruby versions.

georgebrock:
This commit drops support for Ruby 2.0, 2.1, and 2.2, by changing the
build system to look for 2.3+, and also updates the relevant documentation.

CI is also updated to drop older Rubies, and test against the current latest
releases of 2.3 to 2.6.


mike-burns:
@mike-burns note to self: email the OpenBSD Ruby maintainer before agreeing to this PR.

georgebrock:
Interesting spec failure on CI with Ruby 2.6: a `SecurityError` from `File.expand_path('../file_system', __FILE__)` in `spec_helper.rb`. I wonder if the default `$SAFE` level has changed …

I'll need to figure that out before merging this.

georgebrock:
@mike-burns Is there a specific Ruby version we should continue supporting for OpenBSD users?

If I'm looking in the right place, the [OpenBSD lang/ruby port][1] seems to have versions from 1.8–2.6, with [2.6 added very recently][2].

[1]: http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/lang/ruby/
[2]: http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/lang/ruby/Makefile?rev=1.90&content-type=text/x-cvsweb-markup

mike-burns:
https://man.openbsd.org/ruby-module.5

> The ports system defaults to using ruby 2.5 if the version of ruby is not specified. If you want to specify a version for a gem port, use a specific `FLAVOR`, such as `ruby24` to use ruby 2.4. If you want to specify a version for a non-gem port, set `MODRUBY_REV` for the version of ruby you want to use (`2.4`, `2.5`, `jruby`).

This should be fine on the OpenBSD side of things. LGTM!

PR #346, opened 2019-01-27
Update supported Ruby binary names.

georgebrock:
This matches the versions of Ruby we test against on CI, and removes versions we no longer support.

We had trouble with v0.11 on OpenBSD because the binary names we supported weren't up-to-date with the latest Ruby versions (fixed in 428ddb68a07b09ae85edb4f4d9f9ad19e67ce839 and released in v0.11.2). I assume we'll run into this again now that OpenBSD is defaulting to Ruby 2.5.

Probably should have done this as part of #344.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,4 +124,38 @@ gitsh is packaged and installed using GNU autotools.
 
         make release
 
+## Regular maintenance
+
+### Updating supported Ruby versions
+
+When a new version of Ruby is released, or an old version of Ruby reaches
+end-of-life, we should update gitsh's requirements to match.
+
+1. Update the build system (`configure.ac`):
+    - Change the `AX_PROG_RUBY_VERSION` call to the minimum supported version.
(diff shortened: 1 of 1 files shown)


mike-burns:
Ah, thank you! I had forgotten this too. Might help to add a note somewhere (`CONTRIBUTING.md`?).

georgebrock:
Good call. I added descriptions of the various bits of maintenance I've been doing this weekend to `CONTRIBUTING.md`

CONTRIBUTING.md
```diff
@@ -124,4 +124,38 @@ gitsh is packaged and installed using GNU autotools.
...
+
+Gems can be updated in the usual way:
+
+    bundle update
+
+When updating Rubygems for gitsh, there are a few things to consider:
+
+- The Gemfile is only used by developers and maintainers. When gitsh is
+  distributed, all of the gems' files are included in the distribution.
+- Use the minimum supported Ruby version when updating gems, to avoid
```
mike-burns: This comma is optional (and might be wrong, actually).

PR #347, opened 2019-01-27
Set `rl_readline_name` to "gitsh"

### LINKED ISSUE #294: Set rl_readline_name to gitsh
GNU Readline supports conditional configuration in a user's `.inputrc` that targets a specific program. To support this, programs that use Readline set `rl_readline_name`.

Because gitsh's `LineEditor` module is based on Ruby's `Readline` module, we inherited an `rl_readline_name` value of `"Ruby"`. We should change that to `"gitsh"`, add test coverage for gitsh-specific `.inputrc` directives, and document the whole thing in gitsh(1) the manual page.

Readline docs: http://web.mit.edu/gnu/doc/html/rlman_1.html#SEC10

georgebrock:
Fixes #294.

This change allows users to set conditional rules in their `.inputrc` that only apply to gitsh.

See: http://web.mit.edu/gnu/doc/html/rlman_1.html#SEC10

### DIFF
--- a/ext/gitsh/src/line_editor.c
+++ b/ext/gitsh/src/line_editor.c
@@ -1544,7 +1544,7 @@ Init_line_editor_native(void)
     VALUE history, fcomp, ucomp, version;
 
     /* Allow conditional parsing of the ~/.inputrc file. */
-    rl_readline_name = (char *)"Ruby";
+    rl_readline_name = (char *)"gitsh";
 
 #if defined HAVE_RL_GETC_FUNCTION
     /* libedit check rl_getc_function only when rl_initialize() is called, */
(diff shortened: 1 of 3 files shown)


ext/gitsh/src/line_editor.c
```diff
@@ -1544,7 +1544,7 @@ Init_line_editor_native(void)
     VALUE history, fcomp, ucomp, version;
 
     /* Allow conditional parsing of the ~/.inputrc file. */
-    rl_readline_name = (char *)"Ruby";
+    rl_readline_name = (char *)"gitsh";
```
sharplet: Is the cast to `char *` needed?

georgebrock: Probably not, but I've tried to keep the changes from the upstream Ruby `Readline` module to a minimum.

Issue #314, opened 2017-08-11
Add more matchers to the tab completion DSL


georgebrock:
Various rules in the completions file use the a less specific matcher to represent a specific value. In many cases we use the `$anything` matcher, e.g.

```
log $opt*
  --tags $anything?
```

These could be replaced with more specific matchers, e.g.

```
log $opt*
  --tags $tag?
```

The following things could be useful:

- [ ] `$stash` for the names of Git stashes, e.g. `stash show $stash`
- [ ] `$tag` for the names of Git tags, e.g. `log --tags $tag`
- [ ] `$dir_path` for paths that should always be a directory not a file, e.g. `:cd $dir_path`
- [ ] `$branch` for revisions that should always be a branch name, e.g. `branch --delete $branch`
- [ ] `$var_name` for the names of gitsh variables, e.g. `:set $var_name $anything`
- [ ] `$config_var_name` for the names of Git config variables, e.g. `config $config_var_name`
- [ ] `$author` for names or emails of contributors to the repo, e.g. `log --author $author`
- [ ] `$modified_path` for paths that have uncommitted changes, e.g. `add $modified_path`
- [ ] `$ctag` for program symbols defined in a ctags(1) file, e.g. `log -S $ctag`

benknoble:
Not sure if this is the right place to ask, but could `gitsh` also support completing the single- and double-dash option names? I get this in bash (can't seem to pin down from where though--could be the git completion that ships with homebrew git, or from homebrew bash-complete).

georgebrock:
@benknoble The currently released version doesn't, but the master branch does. It will be included in the next release.

georgebrock:
I think `$branch` and `$tag` were the most important items on this list. Now they're done, I'm going to move this issue from v0.14 to v1.0.

PR #353, opened 2019-02-21
Add a $branch variable to tab completion

### LINKED ISSUE #352: branch -D stops completing after second branch name
gitsh version 0.13

I'm trying to clean up my stale local branches, and noticed that after gitsh completes the second branch name, it won't offer any more branch names as completion options.

georgebrock:
Fixes #352

The tab completion DSL had a `$revision` variable which produces branch names, tag names, etc. In some contexts it's too general, e.g. `branch -D` accepts a branch name but not a tag. In those contexts we were using `$anything` to avoid producing invalid tab completion options.

This PR introduces a more specific `$branch` variable that can be used when we want to only tab complete branch names.

### DIFF
--- a/etc/completions
+++ b/etc/completions
@@ -36,7 +36,6 @@ blame $opt* $revision? --? $path
   --porcelain
   --progress
   --reverse $revision
-  --reverse $revision
   --root
   --score-debug
   --show-email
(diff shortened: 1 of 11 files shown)


georgebrock:
Checks off one of the items on #314

PR #354, opened 2019-03-02
Introduce $tag matcher for tab completion.

georgebrock:
The `$tag` matchers produces tag names, and can be used in places where the `$revision` matcher is too permissive, e.g. `tag -d` only accepts a tag name, not a branch name etc.

Checks off one of the items in #314

### DIFF
--- a/spec/units/tab_completion/matchers/tag_matchers_spec.rb
+++ b/spec/units/tab_completion/matchers/tag_matchers_spec.rb
@@ -1,0 +1,65 @@
+require 'spec_helper'
+require 'gitsh/tab_completion/matchers/tag_matcher'
+
+describe Gitsh::TabCompletion::Matchers::TagMatcher do
+  describe '#match?' do
+    it 'always returns true' do
+      matcher = described_class.new(double(:env))
+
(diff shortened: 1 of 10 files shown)


spec/units/tab_completion/matchers/tag_matchers_spec.rb
```diff
@@ -0,0 +1,65 @@
+require 'spec_helper'
+require 'gitsh/tab_completion/matchers/tag_matcher'
+
+describe Gitsh::TabCompletion::Matchers::TagMatcher do
+  describe '#match?' do
+    it 'always returns true' do
+      matcher = described_class.new(double(:env))
+
+      expect(matcher.match?('foo')).to be_truthy
```
mike-burns: Does

```ruby
expect(matcher).to match('foo')
```

work?

georgebrock: Sadly not. I think it would if the method was named differently, but there is already a [`match` matcher][1] in RSpec that takes precedence over the dynamic predicate matchers.

[1]: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/match-matcher

Issue #316, opened 2017-08-11
Improve completions file error handling


georgebrock:
When a completions file is invalid, it's not very clear what went wrong or how to fix it. To make matters worse, completions aren't loaded until the first time the user tries to tab complete something.

This will become even more important with user-specific completions files (#315).

Specific things we should handle are:

- [x] Load completion file(s) on startup instead of first use
- [x] Better error messages for parse errors (e.g. `???`)
- [x] Better error messages for invalid variables (e.g. `$foo`)
- [x] Handling empty files gracefully
- [x] Disallowing `$opt` to be used in the definition of an option, because it causes infinite recursion (e.g. `  --example $opt`)

georgebrock:
There are now PRs open for all of the issues identified here.

PR #355, opened 2019-03-03
Completion config file error handling

georgebrock:
This PR adds error reporting for problems with the tab completion. It covers several of the issues raised in #316.

Specifically:

- Load the completion files on start up, so that errors happen immediately and not when the user first tries to use completion.
- Output useful errors when the file fails to parse (i.e. valid tokens in an invalid order, like `??`).
- Output useful errors when the file contains an invalid variable (e.g. `$foo`)

### DIFF
--- a/lib/gitsh/tab_completion/dsl.rb
+++ b/lib/gitsh/tab_completion/dsl.rb
@@ -6,7 +6,7 @@ module Gitsh
     module DSL
       def self.load(path, start_state, env)
         source = File.read(path)
-        tokens = Lexer.lex(source)
+        tokens = Lexer.lex(source, path)
         factory = Parser.parse(tokens, gitsh_env: env)
         factory.build(start_state)
       rescue Errno::ENOENT
--- a/lib/gitsh/tab_completion/dsl/parser.rb
+++ b/lib/gitsh/tab_completion/dsl/parser.rb
@@ -2,6 +2,7 @@ require 'rltk'
 require 'gitsh/tab_completion/dsl/choice_factory'
 require 'gitsh/tab_completion/dsl/concatenation_factory'
 require 'gitsh/tab_completion/dsl/fallback_transition_factory'
+require 'gitsh/tab_completion/dsl/parse_error'
 require 'gitsh/tab_completion/dsl/maybe_operation_factory'
 require 'gitsh/tab_completion/dsl/null_factory'
 require 'gitsh/tab_completion/dsl/option_transition_factory'
@@ -74,6 +75,8 @@ module Gitsh
             tokens,
             opts.merge(env: Environment.new(opts.fetch(:gitsh_env))),
           )
+        rescue RLTK::NotInLanguage => error
+          raise ParseError.new('Unexpected', error.current)
         end
 
         production(:rule_set) do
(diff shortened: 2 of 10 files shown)


spec/units/tab_completion/dsl/parse_error_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+        RLTK::Token,
+        type: :MAYBE,
+        position: position,
+        value: nil,
+      )
+      exception = described_class.new('Unexpected', token)
+
+      expect(exception.to_s).to eq(
+        'Tab completion configuration error: Unexpected operator (?) '\
+        'at line 2, column 4 in file example.txt'
```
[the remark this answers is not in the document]
georgebrock: There could only ever be one argument in this case

spec/units/input_strategies/interactive_spec.rb
```diff
@@ -165,12 +165,13 @@ def line_editor
 
   def env
     @env ||= double('Environment', {
+      config_directory: '/tmp/gitsh/',
```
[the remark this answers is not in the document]
georgebrock: Disagree

spec/integration/completion_errors_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  context 'with invalid variables' do
+    it 'fails to start up in interactive mode' do
+      with_a_temporary_home_directory do |home|
+        config_path = "#{home}/.gitsh_completions"
+        write_file(config_path, 'add $path $invalid_variable')
+
+        expect { starting_gitsh }.to raise_exception(
+          Gitsh::TabCompletion::DSL::ParseError,
+          /Invalid variable \(\$invalid_variable\) at line 1, column 11 in file #{config_path}/,
```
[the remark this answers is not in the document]
georgebrock: These are too long, but I can't think of a nice way of breaking them up without reducing readability.

lib/gitsh/tab_completion/dsl.rb
```diff
@@ -6,7 +6,7 @@ module TabCompletion
     module DSL
       def self.load(path, start_state, env)
         source = File.read(path)
-        tokens = Lexer.lex(source)
+        tokens = Lexer.lex(source, path)
```
mike-burns: Ah, nice.

lib/gitsh/tab_completion/dsl/parser.rb
```diff
@@ -72,6 +78,8 @@ def self.parse(tokens, opts = {})
             tokens,
             opts.merge(env: Environment.new(opts.fetch(:gitsh_env))),
           )
+        rescue RLTK::NotInLanguage => error
```
mike-burns: Do we want to handle [any of the other three exceptions that RLTK can raise](https://github.com/chriswailes/RLTK/blob/ffb471f52b0a1e12f05927e7e78920727ba4620a/README.md#parsing-exceptions)?

(Maybe not `InternalParserError`.)

georgebrock: I don't think we need to:

- `RLTK::BadToken` should only happen if our lexer and parser disagree on what the set of tokens is, and I'd hope that would be caught by an integration test rather than raising at runtime.
- `RLTK::HandledError` should only happen if the parser uses [error productions][1], which currently we don't, not least because the RLTK README warns that they're not well tested, and because I don't fully understand what they're for.

[1]: https://github.com/chriswailes/RLTK/blob/ffb471f52b0a1e12f05927e7e78920727ba4620a/README.md#error-productions

mike-burns: Excellent reasons. LGTM!

PR #367, opened 2019-04-23
Revision argument to git-reset is optional

sharplet:
`git reset` without specifying a revision defaults to using `HEAD`.

I guessed at the syntax based on the examples nearby.


georgebrock:
@sharplet Looks good. Thanks!

If you're making future changes to this file and want more detail on the format, checkout the gitsh_completions(5) manual page.

PR #389, opened 2024-04-07
fix build for recent... macos? readline? ruby?

jjb:
got these errors on macos 14 with ruby 3.2 and 3.3 and readline 8.2, not sure which one is the culprit, my C skills are almost nonexistent but i powered through!

```
src/line_editor.c:1622:5: error: incompatible function pointer types passing 'VALUE (VALUE, VALUE)' (aka 'unsigned long (unsigned long, unsigned long)') to parameter of type 'VALUE (*)(VALUE)' (aka 'unsigned long (*)(unsigned long)') [-Wincompatible-function-pointer-types]
    rb_define_singleton_method(mLineEditor, "completer_word_break_characters",
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/line_editor.c:1626:5: error: incompatible function pointer types passing 'VALUE (VALUE, VALUE)' (aka 'unsigned long (unsigned long, unsigned long)') to parameter of type 'VALUE (*)(VALUE)' (aka 'unsigned long (*)(unsigned long)') [-Wincompatible-function-pointer-types]
    rb_define_singleton_method(mLineEditor, "completer_quote_characters",
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ GEM
   specs:
     coderay (1.1.2)
     diff-lcs (1.3)
-    ffi (1.10.0)
+    ffi (1.17.0)
     filigree (0.3.3)
     method_source (0.9.2)
     pry (0.12.2)
@@ -33,3 +33,6 @@ DEPENDENCIES
   pry
   rltk
   rspec
+
+BUNDLED WITH
+   2.5.23
--- a/ext/gitsh/src/line_editor.c
+++ b/ext/gitsh/src/line_editor.c
@@ -1164,7 +1164,7 @@ readline_s_set_completer_word_break_characters(VALUE self, VALUE str)
  * Raises NotImplementedError if the using readline library does not support.
  */
 static VALUE
-readline_s_get_completer_word_break_characters(VALUE self, VALUE str)
+readline_s_get_completer_word_break_characters(VALUE str)
 {
     if (rl_completer_word_break_characters == NULL)
         return Qnil;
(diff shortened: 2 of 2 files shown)


jjb:
i have a fork with this fix applied, and instructions for a hacky install: https://github.com/jjb/gitsh/blob/main/README-hacky-install.md

jjb:
i think something like this works

```bash
git checkout jjb-1
bundle --redownload
RUBY=$(which ruby) ./configure
make
sudo make install
```

pjmorse:
I'm not having any luck with @jjb 's 17 November 2024 recipe. I fail on the third step; I get `-bash: ./configure: No such file or directory`. I tried changing that line to `RUBY=$(which ruby) ./configure.rb` (after `chmod 744 ./configure.rb`) but that has its own problems:

```
./configure.rb: line 1: require: command not found
./configure.rb: line 3: syntax error near unexpected token `('
./configure.rb: line 3: `VAR_PATTERN = /\$\((?<name>[a-z_]+)\)/i'
```
Am I missing a step somewhere?

jjb:
“ require: command not found” seems like a ruby script is being run as a shell script

pjmorse:
It does, doesn't it? But the earlier `./configure: No such file or directory` suggests that there's a pre-Makefile step which needs to happen. Looking at commit messages, it seems like the `configure.rb` file is not that. But the `INSTALL` document is _very_ clear that `./configure` needs to happen without providing any clarity about _how_. (Also, there's a strong bias for "system rubies" rather than e.g. RVM-installed Ruby, which is kind of a problem when the current MacOS Ruby is 2.6.10.)

